PR #4, opened 2023-12-18
Clean up build and CI

johnbelamaric:
This PR moves code from the GoogleContainerTools/kpt modules to nephio-project/porch modules. Because there are "internal" packages used, those now live here in nephio-project.

Unfortunately, the api/kptfile/v1 code bleeds internal package types by using them as parameters for public APIs in some places. This means that I had to copy those packages as well, at least until we fix the upstream kpt packages to not leak internal package types.

I also simplifed the Dockerfiles for several builds, as they were unnecessarily complex, leading to build failures.

Lastly, since some of the controllers that we do not need would have pulled in even more code from kpt, I just deleted them.


johnbelamaric:
@radoslawc it seems there is some additional Prow config I need to get presubmits / CI working in this repo, can you take a look?

johnbelamaric:
Ok, so, I am passing all Porch e2e tests with images build from this PR.

To get the Porch e2e CI working, we need the following:

1. Build and push images from the PR (`make push-images` should do it, but I get permission failures trying to do that. I can `make build-images` and then `docker push` but something is weird in the way the Makefile pushes).
2. Create a cluster and deploy those images to it (`make deploy`), and wait for porch to come up.
3. Run the e2e tests with `make test-porch`.
4. If it fails, make the output available somewhere.
5. Clean up the cluster.

@gvbalaji @radoslawc can I get help with the above?

radoslawc:
@johnbelamaric I've submitted PR to enable presubmit jobs for that repo. For image builds top level Makefile defines 4 images:
```
PORCH_SERVER_IMAGE ?= porch-server
PORCH_FUNCTION_RUNNER_IMAGE ?= porch-function-runner
PORCH_CONTROLLERS_IMAGE ?= porch-controllers
PORCH_WRAPPER_SERVER_IMAGE ?= porch-wrapper-server
```
As well as there are:
```
kube-apiserver
etcd
```
in build directory. Are those image names that they should be pushed to docker hub or do we differentiate them somehow (we already have images called porch-wrapper-server etc.)?

johnbelamaric:
The PORCH_* image builds are moving here, these will replace the ones from the kpt project (which is giving us Porch and will eventually remove it from kpt once we are ready).

If no IMAGE_TAG is set, the Makefile currently tags this based on the USER and the git commit. My thinking is that the CI runner can have a username like "nephio-ci", and so whenever there is a PR it will build these images and push them with the tag based on that name and the git commit. We will use that tag in the e2e run.

johnbelamaric:
/test all

I expect failures, because we don't set up the cluster yet for the e2e. But I want to see what happens...

PR #8, opened 2023-12-19
Revise caching strategy

johnbelamaric:
This contains the changes from https://github.com/kptdev/kpt/pull/4094.

It needs more e2e tests though before merging.

### DIFF
--- a/pkg/cache/draft.go
+++ b/pkg/cache/draft.go
@@ -16,6 +16,7 @@ package cache
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/nephio-project/porch/pkg/repository"
 )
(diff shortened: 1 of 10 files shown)


johnbelamaric:
/hold

This will pass the Porch E2E tests, but seems to still have some issues that need fixing. I need to create some new E2E tests that will fail due to those issues, and debug this before we merge it.

johnbelamaric:
Umm. I don't remember the details from back in December. Our TNA team identified some issues in their testing so we backed this change out. @tliron can you reach out to the team and they can perhaps provide some insight?

liamfallon:
This PR introduces a separate mutex for reconciling the cache, which does not block reads. Therefore reads can proceed even if there is a reconcile ongoing as long as the reconcile is not actually updating the cache or the repo. It also consolidates the cache recnciliation into a  new method, which does a 3-way reconciliation against the cache, what's in the CRDs in etc and what's in Git.

The implementation in this PR was merged manually into the current main and the performance of Porch with this PR compared with the current main is shown on the plot below. The change introduced in https://github.com/nephio-project/porch/pull/129 increases the performance of Porch much more so than this PR does.

Therefore I suggest we close this PR.



Screenshot 2024-11-11 at 14 41 13

liamfallon:
Closed, we are adopting a different strategy for caching.

PR #9, opened 2023-12-19
Fix problem with git subdir repos

johnbelamaric:
Fixes https://github.com/nephio-project/nephio/issues/451


johnbelamaric:
I manually ran the Porch e2e tests successfully with images build from this PR.

PR #43, opened 2024-05-08
Refactor of the Developer Environment setup script

kispaljr:
Includes the following changes:
- remove edge1 cluster
- ensure that Gitea's nodePort matches with the port that is forwarded to the host in the kind cluster's config
- the following values are turned into parameters: 
  - name of kind cluster, 
  - name of git repo, 
  - Gitea external IP
- script made idempotent (you can rerun it without any cleanup)
- script exits on error
- Gitea install package is manipulated by KRM functions instead of sed
- check if git user.name config is set
- bugfix: wait for Deployment availability instead of Pod readiness

### DIFF
--- a/docs/tutorials/porch-development-environment/bin/setup.sh
+++ b/docs/tutorials/porch-development-environment/bin/setup.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#! /bin/bash -e
 
 # Copyright 2024 The kpt and Nephio Authors
 #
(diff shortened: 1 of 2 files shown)


docs/tutorials/porch-development-environment/bin/setup.sh
```diff
@@ -14,93 +14,107 @@
...
+  git config --global user.email "myemail@example.com"
+ENDOFEMSG
+  exit 128
 fi
 
-# Create mgmt and edge1 clusters in kind
-curl -s https://raw.githubusercontent.com/nephio-project/porch/main/docs/tutorials/starting-with-porch/kind_management_cluster.yaml | \
-  kind create cluster --config=-
-
-curl -s https://raw.githubusercontent.com/nephio-project/porch/main/docs/tutorials/starting-with-porch/kind_edge1_cluster.yaml | \
```
efiacor, line 29: Not a major issue but I think the orig guide, which this is an extension of is using the egde1 cluster.
https://github.com/nephio-project/porch/tree/main/docs/tutorials/starting-with-porch#deploying-a-blueprint-onto-a-workload-cluster

kispaljr: Yes, I was a bit uncertain regarding removing edge1. My rationale was that the "Starting with Porch" tutorial is more like a "User's Guide", while this script belongs to the "Setup the Developer Environment" guide. And I think edge1 is not needed to develop and test porch, it belongs more to ConfigSync. 
All-in-all I can add back edge1 if we think there is need for it during development.

efiacor: Ye I guess either way the README guides will need to be updated a bit.
May be best to get @liamfallon direction on it when he is back on 13/05

liamfallon (commented):
This looks great! Thanks for cleaning up the setup @kispaljr

PR #45, opened 2024-05-11
Add tilt for debug configuration

nagygergo:
Added dockerfiles, scripts and docs on how to do live debugging of the various porch components. 
The wrapper server and the kpt function processes are not easily modifiable this way, so those were left out. 
If there is need to debug into those, then the attached can be extended.


docs/tutorials/porch-development-environment/debugging.md
```diff
@@ -0,0 +1,69 @@
...
+As a first step, in the `tilt_config.json` file, the correspondig entry should be set to true.
+Then on the web UI, the reload button should be pressed for the respective service.
+When the image with the debug enabled is started up, the container will log something like the following:
+`info layer=debugger launching process with args:`. This means the debugger wrapper is up, and waiting for a connection. 
+At this point, open VS code, and select the "Remote Porch Server", "Remote Porch Controllers" or "Remote Porch Function Runner", depending on the component.
+This will attach the VS code instance to the container, and the container process will start up. It's possible to debug startup issues this way, as the debugger server won't start the porch processes until there is a debugger client (VS Code) attached to it.
+
+### What's happening underneath
+For each of the porch components there is a `Dockerfile.debug`, which has 2 significant differences from the original process.
+The application code is compiled with flags so that all inlining of code and other optimalizations done by the compiler are disabled, so the debugger can track source code lines correctly.
```
efiacor, line 67:
```suggestion
The application code is compiled with flags so that all inlining of code and other optimizations done by the compiler are disabled, so the debugger can track source code lines correctly.
```

docs/tutorials/porch-development-environment/debugging.md
```diff
@@ -0,0 +1,69 @@
...
+A wrapper application is replacing the entrypoint of the container, called [delve](https://github.com/go-delve/delve). This will expose a rpc API called [dap](https://microsoft.github.io/debug-adapter-protocol/overview), which can be used by many IDEs for remote debugging. The ports `:4000`, `:4001`, `:4002` are hardcoded into the respective images. Delve will wait until there is a debugger attached before kicking off the porch application process.
```
efiacor, line 68:
```suggestion
A wrapper application is replacing the entrypoint of the container, called [delve](https://github.com/go-delve/delve). This will expose an rpc API called [dap](https://microsoft.github.io/debug-adapter-protocol/overview), which can be used by many IDEs for remote debugging. The ports `:4000`, `:4001`, `:4002` are hardcoded into the respective images. Delve will wait until there is a debugger attached before kicking off the porch application process.
```

docs/tutorials/porch-development-environment/debugging.md
```diff
@@ -0,0 +1,69 @@
...
+There is a configuration in `Tiltfile`, which whenever one of the containers are deployed in debug mode, creates a `kubectl port-forward` for the respective port to `localhost`. This means that VS Code can now attach it's remote debugger to the `localhost:400x` port, and start monitoring the running application in k8s.
```
efiacor, line 69:
```suggestion
There is also configuration in `Tiltfile`, which whenever one of the containers are deployed in debug mode, creates a `kubectl port-forward` for the respective port to `localhost`. This means that VS Code can now attach it's remote debugger to the `localhost:400x` port, and start monitoring the running application in k8s.
```

Tiltfile
```diff
@@ -0,0 +1,79 @@
...
+cfg = config.parse()
+
+# There are cross-dependencies, so whenever a single file is changed, everything would be redeployed with TRIGGER_MODE_AUTO
+trigger_mode(TRIGGER_MODE_MANUAL)
+
+#TODO(nagygergo): Discover these automatically?
+crds = [
+    "api/porchconfig/v1alpha1/config.porch.kpt.dev_functions.yaml",
+    "api/porchconfig/v1alpha1/config.porch.kpt.dev_repositories.yaml",
+    "controllers/config/crd/bases/config.porch.kpt.dev_fleetmembershipbindings.yaml",
```
efiacor, line 14: The fleet stuff is junk left over from the migration from kpt and needs to be removed/moved to another repo. We don't use it so can be ignored. Same goes for the Function CRD above.

deployments/porch/2-function-runner.yaml
```diff
@@ -37,16 +37,15 @@ spec:
       serviceAccountName: porch-fn-runner
       containers:
         - name: function-runner
-          image: gcr.io/example-google-project-id/porch-function-runner:latest
+          image: nephio-project/porch-function-runner:latest
```
efiacor, line 40: Not sure if this might cause issues with some of the other Make targets.

deployments/porch/9-controllers.yaml
```diff
@@ -41,7 +41,9 @@ spec:
       containers:
       - name: porch-controllers
         # Update to the image of your porch-controllers build.
-        image: gcr.io/example-google-project-id/porch-controllers:latest
+        image: nephio-project/porch-controllers:latest
+        args:
+          - --reconcilers=*
```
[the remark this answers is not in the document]
nagygergo: Yes, this package is missing an ClusterRole/RoleBinding for the porch-controllers as well. Do you think if it's a good idea to manage the deployment artifacts outside the repo? Or you want to copy it into this repo?

deployments/porch/2-function-runner.yaml
nagygergo: Seems like `scripts/create-deployment-blueprint.sh` is looking for those, for replacing them. Fixed it, but I couldn't get the associated make target working, seems like has some google things in there.

docs/tutorials/porch-development-environment/debugging.md
nagygergo: Done

docs/tutorials/porch-development-environment/debugging.md
nagygergo: Done

docs/tutorials/porch-development-environment/debugging.md
nagygergo: Done

deployments/porch/2-function-runner.yaml
efiacor: Ye I think some/most of that google specific stuff should be removed.
I started using the same pkg when doing the kind deploy target but switched to use the catalog pkg instead.
https://github.com/nephio-project/porch/blob/main/Makefile#L325

We need to refine the Make file to reduce the bloat.

deployments/porch/9-controllers.yaml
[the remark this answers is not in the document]
efiacor: It's a messy one, but ideally we have 1 source of truth for the porch pkg. The official repo is the catalog repo.
It does break/confuse the Make/test setup somewhat though.

liamfallon:
@nagygergo Thanks for the great work.

I think we should not integrate Tilt into the main codebase but should place this work in the Porch developer tutorial (or a separate tutorial) so that users can set Tilt and in-cluster debugging themselves, as @tliron suggested. If we integrate Tilt supprot into the main codebase, then we will have to provide ongoing support for it and also offer support for any environment for all comers. This is likely to be very difficult to do.

If we provide the description of how to set Tilt up and provide example configurations (stating the environment on which the configurations worked), then we can ask users to download and tweak those configurations themselves. I think this is reasonable because any developer that is attempting in-cluster debugging will likely need to be an advanced user and should be capable of tweaking the config.

liamfallon (commented):
See comment above.

nagygergo:
The current make targets do most of the things that are in this one. No need to have this open anymore

PR #44, opened 2024-05-08
Developer Env Setup script: add more automation steps

kispaljr:
This depends on #43 

Automate even more steps of the tutorial: https://github.com/nephio-project/porch/tree/main/docs/tutorials/porch-development-environment.

The new steps added are detailed in the updated [README.md](https://github.com/nokia/nephio-porch/tree/dev-env-setup-cert-gen/docs/tutorials/porch-development-environment) file.

Please NOTE: Contrary to what the documentation says the end-to-end test will fail after executing the setup script. It is a known fact, and #57 was created to fix it.

PR #57, opened 2024-05-25
Make end-to-end tests work with a local porch-server

kispaljr:
Right now some end-to-end test cases fail if the porch-server is running locally and not in-cluster. (The same tests are working properly, if all components are running in-cluster)

This PR intends to fix this problem. 
Fixes https://github.com/nephio-project/nephio/issues/743

End-to-end test can be run with the following commands:
```
cd <porch-repo>
./docs/tutorials/porch-development-environment/bin/setup.sh
E2E=1 go test -v ./test/e2e
E2E=1 go test -v ./test/e2e/cli/
```

The above commands should all succeed with the code in this PR.

This PR is based on the loosely related #44


liamfallon (commented):
On a read through this looks good. I'll download the PR and try it in my environment.

kispaljr:
NOTE: I've added instructions to run the end-to-end tests to the description of the PR.

.vscode/launch.json
```diff
@@ -23,14 +12,30 @@
...
                 "--secure-port=4443",
-                "--v=7",
-                "--standalone-debug-mode",
+                // "--v=7",
+                // "--standalone-debug-mode",
                 "--kubeconfig=${env:KUBECONFIG}",
-                "--cache-directory=${workspaceFolder}/.cache",
                 "--function-runner=172.18.255.201:9445",
-                "--repo-sync-frequency=60s"
+                "--repo-sync-frequency=60s",
```
nagygergo, line 19: Is jsonlint ok with this one?

.vscode/launch.json
```diff
@@ -23,14 +12,30 @@
             "program": "${workspaceFolder}/cmd/porch/main.go",
             "args": [
                 "--secure-port=4443",
-                "--v=7",
-                "--standalone-debug-mode",
+                // "--v=7",
+                // "--standalone-debug-mode",
                 "--kubeconfig=${env:KUBECONFIG}",
-                "--cache-directory=${workspaceFolder}/.cache",
```
nagygergo, line 29: Removing this will mean it will work into /home/<userid>/.cache, or something similar. Might be an unintended sideeffect.

build/Dockerfile.apiserver
```diff
@@ -15,14 +15,13 @@
 FROM golang:1.22.2-bookworm as builder
 
 WORKDIR /workspace/src
-RUN git clone https://github.com/kubernetes/kubernetes --branch v1.23.2 --depth=1
+RUN git clone https://github.com/kubernetes/kubernetes --branch v1.30.1 --depth=1
```
nagygergo, line 18: Any ideas why the env needs to rebuild it instead of using registry.k8s.io/kube-apiserver:v1.30.1 ?

pkg/apiserver/webhooks.go
```diff
@@ -47,29 +48,76 @@ import (
...
+	Host             string // only used if Type == WebhookTypeUrl
+	Path             string
+	Port             int32
+	CertStorageDir   string
+}
+
+func NewWebhookConfig() *WebhookConfig {
+	var cfg WebhookConfig
+	// NOTE: CERT_NAMESPACE is supported for backward compatibility.
+	// TODO: We may consider using only WEBHOOK_SERVICE_NAMESPACE instead.
```
nagygergo, line 75: Could you write an issue for removing it?
Also is this the right way of issuing certs? I mean tls in not mandatory, the webhooks can be created with insecrueSkipTlsVerify as well for dev envs. 
For production envs, I think using an external cert management entity would be better (see #53 ). For dev envs, could it be just disabled?

scripts/clean-kind-only-e2e-test.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
nagygergo, line 1: Could you create a make target for this script?

test/e2e/e2e_test.go
```diff
@@ -1225,42 +1231,49 @@ func (t *PorchSuite) TestDeleteFromMain(ctx context.Context) {
...
-
+	t.Log("Wait for the 'main' revisions to get created")
 	var firstPkgRevFromMain porchapi.PackageRevision
 	var secondPkgRevFromMain porchapi.PackageRevision
-
-	for _, pkgrev := range list.Items {
-		if pkgrev.Spec.PackageName == packageNameFirst && pkgrev.Spec.Revision == "main" {
-			firstPkgRevFromMain = pkgrev
+	var list porchapi.PackageRevisionList
+	finalDeadline := time.Now().Add(5 * time.Minute)
```
nagygergo, line 1238: Could this be fulfilled by the helper function waitUntilMainBranchPackageRevisionExists?

test/e2e/e2e_test.go
```diff
@@ -1225,42 +1231,49 @@ func (t *PorchSuite) TestDeleteFromMain(ctx context.Context) {
...
 	t.DeleteE(ctx, &porchapi.PackageRevision{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: t.namespace,
 			Name:      firstPkgRevFromMain.Name,
 		},
 	})
 
-	// We need to wait for the sync
-	time.Sleep(75 * time.Second)
+	// t.Log("Wait for the sync")
```
nagygergo, line 1273: Remove?

test/e2e/suite.go
```diff
@@ -167,32 +166,33 @@ func (t *TestSuite) IsUsingDevPorch() bool {
 		Name:      porch.Spec.Service.Name,
 	}, &service)
 
-	return service.Spec.Type == coreapi.ServiceTypeExternalName && service.Spec.ExternalName == "host.docker.internal"
+	isPorchLocal := len(service.Spec.Selector) == 0
+	areAllLocal := service.Spec.Type == coreapi.ServiceTypeExternalName
```
nagygergo, line 170: What all means in this context?

.vscode/launch.json
kispaljr: you are right, this is not nice. Fixing it.

.vscode/launch.json
kispaljr: Yes, you guessed correctly :) removing this was a result of an experimentation that is not relevant to the PR. Fixing this.

build/Dockerfile.apiserver
kispaljr: No, I don't see any reason for re-building . Your suggestion makes more sense to me than the current process. 
Is it OK, if I leave it like this in this PR, and you send a separate PR for your proposal?

scripts/clean-kind-only-e2e-test.sh
kispaljr: sure. I think it will require changes in the way we are calculating the default value of KUBECONFIG in the global Makefile, however that is overdue, anyway.
Btw, I believe we should do a wide-scope clean up of our make targets together as a team. But that is out-of-scope of this PR.

test/e2e/e2e_test.go
kispaljr: I haven't looked into it yet, but just judging by the fact that you proposing it and by the name of the function I am pretty sure it can :)
Thanks, I will fix it.

test/e2e/e2e_test.go
kispaljr: yes, good catch. I mean to do that.  Fixing it.

test/e2e/suite.go
kispaljr: function-runner and the porch-controllers.
However now that you pointed this out. I will replace the "all" part by properly checking if function-runner is running in-cluster or not, because I think that is how the result is actually used in the end-to-end tests.
Thanks.

pkg/apiserver/webhooks.go
kispaljr: I thought about this, and decided to leave this question to SIG Security :) and out-of-scope of this PR. the goal of this PR is to make the end-to-end tests work in the developement environment, and I think that is achieved.

to be honest, since you have to have the proper access to create a ValidationWebhookConfiguration object, it seems fine to me if the CA bundle in it generated on-the-fly, since it is only used for the communication between the kube-apiserver and porch. but again, I am not a security expert so I will let the experts decide :) and #53 will iron out the security aspect

test/e2e/suite.go
kispaljr: done.

PR #69, opened 2024-06-12
Adding porch image builds for every PR merged tagged latest

radoslawc:
Porch images will be built every merged PR and pushed to Docker Hub with tags: 'latest' and usual tag that's prow's job ID


liamfallon (commented):
/approve

PR #70, opened 2024-06-12
Add support for new field selectors: spec.workspaceName and spec.lifecycle

kispaljr:
Fixes https://github.com/nephio-project/nephio/issues/724


liamfallon (commented):
I don't know if it's normal in the Kubernetes world but it might be useful sometime to implement case insensitive field selection.

PR #75, opened 2024-06-20
Expose kpt package rendering errors via the status of PackageVariant

kispaljr:
Fixes https://github.com/nephio-project/nephio/issues/730

This solution exposes the `RenderStatus` of downstream package revisions as Conditions in the status of PackageVariant.


kispaljr:
The reason behind exposing the RenderStatus as a Conditions is that it doesn't require to modify the CRD/schema days before Release date.

If CRD modifications are allowed, then I would consider adding a renderStatus field to items of status.downstreamTargets as a more elegant solution.

kispaljr:
I rewrote the PR to expose render status via `status.downstreamTargets`. As I commented above I consider this a superior solution, even though it requires a minor change in the PC schema.

PR #76, opened 2024-06-20
PackageVariant controller: Use the default exponential back-off in case of errors, instead of a constant 30 sec retry

kispaljr:
This improves the response times of recoverable PackageVariant errors, e.g. normal errors related to optimistic locking when updating PackageRevisionResources objects.

This should improve, but not solve https://github.com/nephio-project/nephio/issues/770 and https://github.com/nephio-project/nephio/issues/462 .


liamfallon (commented):
/lgtm
/approve

PR #77, opened 2024-06-20
Porchctl dev release workflow

radoslawc:
This workflow will build porchctl binary from 'main' branch with just 'make porchctl' each successful merge. This binary whill then be used for e2e tests running 'main'.
Workflow will create release called 'Porchctl build from main' and replace it each build so it's always latest.


liamfallon (commented):
/approve
/lgtm

PR #92, opened 2024-08-13
fix porchсtl push method for windows

RomanBudnyk:
we are compiling the porchctl to use it on Windows. the proposed fix is changing the Windows path representation to the Unix-based one.

### DIFF
--- a/pkg/cli/commands/rpkg/push/command.go
+++ b/pkg/cli/commands/rpkg/push/command.go
@@ -254,7 +254,12 @@ func readFromDir(dir string) (map[string]string, error) {
 	}); err != nil {
 		return nil, err
 	}
-	return resources, nil
+	newRes := make(map[string]string)
+	for key, value := range resources {
+		newKey := strings.ReplaceAll(key, "\\", "/")
+		newRes[newKey] = value
+	}
+	return newRes, nil
 }
 
 func readFromReader(in io.Reader) (map[string]string, error) {
--- a/pkg/cli/commands/rpkg/push/command.go
+++ b/pkg/cli/commands/rpkg/push/command.go
@@ -256,7 +256,7 @@ func readFromDir(dir string) (map[string]string, error) {
 	}
 	newRes := make(map[string]string)
 	for key, value := range resources {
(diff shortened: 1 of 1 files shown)


pkg/cli/commands/rpkg/push/command.go
```diff
@@ -254,7 +254,12 @@ func readFromDir(dir string) (map[string]string, error) {
 	}); err != nil {
 		return nil, err
 	}
-	return resources, nil
+	newRes := make(map[string]string)
+	for key, value := range resources {
+		newKey := strings.ReplaceAll(key, "\\", "/")
```
kispaljr, line 259: Using something like [filepath.ToSlash](https://pkg.go.dev/path/filepath#ToSlash) seems to me a safer alternative to achieve this.

kispaljr: Also, IMHO it makes the code more readable

RomanBudnyk: hi @kispaljr !

thanks for your reply! GO is far away not my primary language, so I am not aware of all the functionality. Let me fix it as you've suggested.

cheers

pkg/cli/commands/rpkg/push/command.go
```diff
@@ -254,7 +254,12 @@ func readFromDir(dir string) (map[string]string, error) {
 	}); err != nil {
 		return nil, err
 	}
-	return resources, nil
+	newRes := make(map[string]string)
+	for key, value := range resources {
```
kispaljr, line 258: I hate being this picky again, sorry in advance :)

Have you considered to do the transformation of the `resources` map keys where the values are added to the map (in line 252) instead of using a second for cycle.
I am thinking of something along the lines of:
`resources[filepath.ToSlash(rel)] = string(contents)`

RomanBudnyk: hi @kispaljr ,

you are absolutely right, done as you suggested. I also hate redundant loops :)
thanks for reviewing!

PR #93, opened 2024-08-22
More sensible defaults for webhook environment variables

kispaljr:
Remove hardwired string literals as the default values of WEBHOOK_SERVICE_NAME and WEBHOOK_SERVICE_NAMESPACE environment variables.

This PR introduces the following changes in behavior:

The value of WEBHOOK_SERVICE_NAME is determined by one of the followings (in precedence order):
- the WEBHOOK_SERVICE_NAME environment variable
- the name of the service specified in the `APIService` object named `v1alpha1.porch.kpt.dev`

The value of WEBHOOK_SERVICE_NAMESPACE is determined by one of the followings (in precedence order):
- the WEBHOOK_SERVICE_NAMESPACE environment variable
- the CERT_NAMESPACE environment variable (deprecated, please don't use this!!!)
- the namespace of the service specified in the `APIService` object named `v1alpha1.porch.kpt.dev`
- namespace of the current process (if running in a pod)

### DIFF
--- a/deployments/porch/5-rbac.yaml
+++ b/deployments/porch/5-rbac.yaml
@@ -38,7 +38,7 @@ rules:
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
   - apiGroups: ["apiregistration.k8s.io"]
     resources: ["apiservices"]
-    verbs: ["get", "watch", "list"]
+    verbs: ["get"]
   # Needed for priority and fairness
   - apiGroups: ["flowcontrol.apiserver.k8s.io"]
     resources: ["flowschemas", "prioritylevelconfigurations"]
--- a/deployments/porch/5-rbac.yaml
+++ b/deployments/porch/5-rbac.yaml
@@ -36,6 +36,9 @@ rules:
   - apiGroups: ["config.porch.kpt.dev"]
     resources: ["packagerevs", "packagerevs/status"]
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  - apiGroups: ["apiregistration.k8s.io"]
+    resources: ["apiservices"]
+    verbs: ["get", "watch", "list"]
   # Needed for priority and fairness
   - apiGroups: ["flowcontrol.apiserver.k8s.io"]
     resources: ["flowschemas", "prioritylevelconfigurations"]
--- a/pkg/apiserver/webhooks.go
+++ b/pkg/apiserver/webhooks.go
@@ -75,8 +75,8 @@ type WebhookConfig struct {
 	CertManWebhook   bool
 }
 
-// NewWebhookConfig creates a new WebhookConfig object filled with values read from environment variables
-func NewWebhookConfig(ctx context.Context) *WebhookConfig {
+// newWebhookConfig creates a new WebhookConfig object filled with values read from environment variables
+func newWebhookConfig(ctx context.Context) *WebhookConfig {
 	var cfg WebhookConfig
 	// NOTE: CERT_NAMESPACE is supported for backward compatibility.
 	// TODO: We may consider using only WEBHOOK_SERVICE_NAMESPACE instead.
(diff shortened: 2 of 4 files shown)


deployments/porch/5-rbac.yaml
```diff
@@ -36,6 +36,9 @@ rules:
   - apiGroups: ["config.porch.kpt.dev"]
     resources: ["packagerevs", "packagerevs/status"]
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  - apiGroups: ["apiregistration.k8s.io"]
+    resources: ["apiservices"]
+    verbs: ["get", "watch", "list"]
```
efiacor, line 41: Do we need all of these?

pkg/apiserver/webhooks.go
```diff
@@ -82,9 +83,48 @@ func NewWebhookConfig() *WebhookConfig {
...
+		if cfg.ServiceName == "" { // empty value and unset envvar are the same for our purposes
+			// if WEBHOOK_SERVICE_NAME is not set, try to use the porch API service name
+			apiSvc, err := util.GetPorchApiServiceKey(ctx)
+			if err != nil {
+				panic(fmt.Sprintf("WEBHOOK_SERVICE_NAME environment variable is not set, and could not find porch APIservice: %v", err))
+			}
+			cfg.ServiceName = apiSvc.Name
+			apiSvcNs = apiSvc.Namespace // cache the namespace value to avoid duplicate calls of GetPorchApiServiceKey()
+		}
+		// the webhook service namespace gets it value from the following sources in order of precedence:
```
efiacor, line 98: Should this go in the docs somewhere instead?

pkg/apiserver/webhooks.go
```diff
@@ -82,9 +83,48 @@ func NewWebhookConfig() *WebhookConfig {
...
-		cfg.ServiceNamespace = getEnv("WEBHOOK_SERVICE_NAMESPACE", "porch-system")
-		cfg.ServiceNamespace = getEnv("CERT_NAMESPACE", cfg.ServiceNamespace)
+
+		var apiSvcNs string
+		cfg.ServiceName = os.Getenv("WEBHOOK_SERVICE_NAME")
+		if cfg.ServiceName == "" { // empty value and unset envvar are the same for our purposes
+			// if WEBHOOK_SERVICE_NAME is not set, try to use the porch API service name
+			apiSvc, err := util.GetPorchApiServiceKey(ctx)
+			if err != nil {
+				panic(fmt.Sprintf("WEBHOOK_SERVICE_NAME environment variable is not set, and could not find porch APIservice: %v", err))
```
efiacor, line 93: Should we try to recover instead of panic? Can we run without the webhook?

pkg/apiserver/webhooks.go
```diff
@@ -82,9 +83,48 @@ func NewWebhookConfig() *WebhookConfig {
...
+			}
+		}
+		if cfg.ServiceNamespace == "" {
+			var err error
+			cfg.ServiceNamespace, err = util.GetInClusterNamespace()
+			if err != nil {
+				// this was our last resort, so panic if failed
+				panic(fmt.Sprintf("WEBHOOK_SERVICE_NAMESPACE environment variable is not set, and could determine in-cluster namespace: %v", err))
+			}
+		}
```
efiacor, line 123: Candidate for a helper function maybe?

deployments/porch/5-rbac.yaml
kispaljr: no. you are right, we only need get. will fix.

pkg/apiserver/webhooks.go
kispaljr: yes. I created an issue: https://github.com/nephio-project/nephio/issues/804

pkg/apiserver/webhooks.go
kispaljr: Without a working webhook nothing prevents actually deleting PackageRevisions that are not in the DeletionProposed state. As such the API server cannot properly run without the webhook, so IMHO it is better to panic here,  than trying to recover.

pkg/apiserver/webhooks.go
kispaljr: fixed.

pkg/apiserver/webhooks.go
```diff
@@ -137,10 +99,56 @@ func NewWebhookConfig(ctx context.Context) *WebhookConfig {
 	return &cfg
 }
 
-var (
-	cert        tls.Certificate
-	certModTime time.Time
-)
+// WebhookServiceName returns the name and namespace of Kubernetes service belonging to the webhook
+func WebhookServiceName(ctx context.Context) (serviceName, serviceNamespace string) {
```
nagygergo, line 103: Could this be a private function?

kispaljr: fixed

PR #95, opened 2024-08-23
 Retry 'Update' operations that are failing due to normal optimistic locking behavior in CLI commands

kispaljr:
We are using the Update API operation in the porchctl CLI commands for multiple purposes. Those can fail due to normal optimistic locking behavior if another process tries to manipulate the same object (typically PackageRevision) in parallel. In a reconciliation loop these naturally occurring errors correctly cause a retry, but in the CLI these errors cause the command to fail. 
This PR adds a retry loop around those Update calls, thus implementing default optimistic locking behavior.
This should fix some flaky errors that we experienced in our Nephio end-to-end tests, e.g.: https://github.com/nephio-project/nephio/issues/803. 

NOTE: using server-side apply instead of Update doesn't work when we try to change the lifecycle state of PackageRevisions, because of at least these two reasons:
- the porch API server doesn't support it
- we have to be sure what the current Lifecycle state of the PackageRevision is before applying the new state. Server-side apply doesn't provide this guarantee, but optimistic locking does.

Fixes https://github.com/nephio-project/nephio/issues/803

### DIFF
--- a/internal/kpt/util/porch/approval.go
+++ b/internal/kpt/util/porch/approval.go
@@ -21,24 +21,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func UpdatePackageRevisionApproval(ctx context.Context, client client.Client, key client.ObjectKey, new v1alpha1.PackageRevisionLifecycle) error {
-
-	var pr v1alpha1.PackageRevision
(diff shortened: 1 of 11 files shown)


kushnaidu (approved):
Looks good to me

liamfallon (commented):
/approve

PR #97, opened 2024-08-27
Prepare adding end-to-end tests to the controller go module

kispaljr:
This PR contains the following end-to-end testing related changes:
- Export end-to-end TestSuite methods as public in order to be usable from other go modules in the future (notable from the porch/controller go module)
- Fix a bug in `make run-in-kind` that prevented some docker bulid errors to properly propagated to the root make process
- Force the reload of the newly built porch-server and/or porch-controllers pods during `make run-in-kind`
- test/e2e/cli now explicitly tests the porchctl binary in the .build directory of the current git tree (as opposed to the first one found in PATH)
- Hopefully make the TestFunctionRepository test case less flaky by changing a fix delay to a more flexible wait cycle
- pin the version of the kpt-functions-sdk/go/fn module to prevent future dependency updates to break the code of the  imported kpt functions that unfortunately use (the pinned) ancient incompatible version of the SDK


Makefile
```diff
@@ -286,35 +286,37 @@ load-images-to-kind: ## Build porch images and load them into a kind cluster
...
 		echo "Skipping building $(IMAGE_REPO)/$(PORCH_FUNCTION_RUNNER_IMAGE):${IMAGE_TAG} as it is already loaded into kind" ; \
 	fi
     # NOTE: SKIP_PORCHSERVER_BUILD must be evaluated at runtime, hence the shell conditional (if) here
 	@if [ "$(SKIP_PORCHSERVER_BUILD)" = "false"	]; then \
 		echo "Building $(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):${IMAGE_TAG}" ; \
-		docker buildx build --load --tag $(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):$(IMAGE_TAG) -f ./build/Dockerfile "$(PORCHDIR)" ; \
-		kind load docker-image $(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):${IMAGE_TAG} -n ${KIND_CONTEXT_NAME} ; \
+		docker buildx build --load --tag $(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):$(IMAGE_TAG) -f ./build/Dockerfile "$(PORCHDIR)" && \
+		kind load docker-image $(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):${IMAGE_TAG} -n ${KIND_CONTEXT_NAME} && \
+		kubectl delete deployment -n porch-system --ignore-not-found=true porch-server ; \
```
efiacor, line 307: Should we use rollout restart here instead?

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/nephio-project/porch
 
 go 1.22
 
+replace github.com/GoogleContainerTools/kpt-functions-sdk/go/fn => github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220506190241-f85503febd54
```
efiacor, line 5: This is a pain. How do we handle this dependency going forward?

test/e2e/suit_utils.go
efiacor, line 1: "test/e2e/suite_utils.go"

Makefile
kispaljr: in short: we don't want to start the new pods at this point

a bit longer: :)
the old process were:
- rebuild the necessary images and load them to kind
- calculate the new porch kpt package (potentially remove server/controller)
- apply the new config  --> here, if, let's say the "porch-server" Deployment object hasn't changed (but there is a docker image for the porch server), then the server wasn't reloaded --> this is the bug we want to fix

the new process:
- rebuild the necessary images and load them to kind,
- --> stop the pods running the docker images that were rebuilt in the previous step (and only those pods). We rely on the upcoming steps to calculate the new Deployment objects (that may be different from the current ones) and to actually apply them to the cluster, and thus eventually starting the new pods
- [no change in the rest of the process]

test/e2e/suit_utils.go
kispaljr: ??
was this comment accidental? :)

kispaljr: I get it now, will fix :)

go.mod
efiacor: Not really so much about the multi module situation but more about the sdk itself. It's pretty much dead right and we rely heavily on it

kispaljr: I like the SDK, I think it is a good project, and I even have some contributions lined-up since Nephio R1. However I agree that it seems to be abandoned by Google, but nobody can take over, or even contribute who is not allowed to sign the Google CLA (like me). It should have been migrated to the kpt CNCF project by now, but that hasn't happened for more than a year. I asked about te migration here: https://github.com/kptdev/kpt/issues/4034, and also asked various Google people, but no news on the ETA.
Btw. the same goes to the kpt function catalog. It seems to be abandoned, and on the kpt slack there is at least one guy who also waits for the migration to be able to contribute. And an abandoned kpt function catalog doesn't help the adoption of neither kpt, nor Nephio.

Maybe we could fork those under Nephio? However I don't think that we have the capacity to actually maintain them.

liamfallon (commented):
/approve
/lgtm

PR #107, opened 2024-09-03
Merge all of porch's go modules into a single go module

kispaljr:
Currently porch consists of three separate go modules:
- the "root" module that contains primarily the  porch server
- the controllers module that contains the packagevariant* controllers
- the api module that contains (some of the) porch api definitions in the form of go structs

These modules depend on each other, and that makes it really hard to introduce changes that touches multiple modules.
This PR aims to put all of the above into one singular porch go module to make the lives of developers considerably easier.


kispaljr:
This is still work-in-progress. I will remove the hold label when it is ready for review.

kispaljr:
@efiacor , @liamfallon this is ready for review now

pkg/engine/builtinruntime.go
```diff
@@ -16,15 +16,16 @@ package engine
...
 	"io"
 
-	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/apply-replacements/replacements"
-	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-namespace/transformer"
-	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/starlark/starlark"
 	fnsdk "github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 	"github.com/nephio-project/porch/pkg/kpt"
 	v1 "github.com/nephio-project/porch/pkg/kpt/api/kptfile/v1"
 	"github.com/nephio-project/porch/pkg/kpt/fn"
+	"github.com/nephio-project/porch/pkg/kpt/fn/third_party/GoogleContainerTools/kpt-functions-catalog/functions/go/apply_replacements"
```
efiacor, line 26: I understand the logic in pulling these in but it concerns me that we then "own" them.
Not quite clear on what these are used for either. Could you explain?

go.mod
```diff
@@ -4,16 +4,10 @@ go 1.22.0
...
-
 require (
 	cloud.google.com/go/iam v1.1.12
 	github.com/GoogleContainerTools/kpt v1.0.0-beta.48
-	github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/apply-replacements v0.1.1
-	github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/apply-setters v0.2.0
-	github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/set-namespace v0.4.1
-	github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/starlark v0.4.3
-	github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220506190241-f85503febd54
+	github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20230427202446-3255accc518d
```
efiacor, line 10: How do we handle this one going forward also?
This is being used in the nephio codebase too.

controllers/Makefile
```diff
@@ -16,15 +16,6 @@ IMAGE_TAG ?= latest
 IMAGE_REPO ?= docker.io/nephio
 IMAGE_NAME ?= porch-controllers
 
-# This includes the following targets:
```
efiacor, line 19: No longer needed?

efiacor (commented):
Looks great. I'm fully in favour of this approach but could this cause other potential issue that I don't have the golang knowledge to foresee?

efiacor (commented):
Also, I didn't fully check but we need to ensure that the code-gen stuff is still sound.

pkg/engine/builtinruntime.go
kispaljr: sure. the problem with KRM (formerly kpt) functions is that actually calling out to function-runner to execute the docker container and swnd back the image can be ridiculously slow for simply things like replacing a namespace. Because of this there are various spped up tricks are buolt into porch to make calling at least know/core functions faster (e.g. the binary of lot of them is actually copied to the function-runner container to avoid actually starting a new pod). One of those speed up measured is to have a so-called "built-in executor" in porch that instead of relying on function-runner calls the golang function for some known KRM functions. 
This require us to import the go code into porch. The problem with this  (as you and me are painfully aware) is that the code of those functions still uses outdated versions of the kpt-fn-sdk that is incompatible with our own PackageVariant code. That's why instead of simply importing it I copied/forked the code from the catalog repo, updated the code to use the latest version of the sdk. And all of this let's us get rid of the dependency on the ancient sdk version. yay :)

Please note that in the future we can think of other solutions, e.g. we can fork the whole catalog, since the old one seems yo be abandoned by Google.

(I also fixed a bug that in case of a panicking KRM function, that is normal behavior on invalid input btw., also panics the whole porch server)

go.mod
kispaljr: now both our code and the 3rd party KRM function code relies on the same version of the sdk, namely the latest.

pkg/engine/builtinruntime.go
efiacor: Much clearer now. Thanks

kispaljr:
> Also, I didn't fully check but we need to ensure that the code-gen stuff is still sound.

I can only state that `make generate` runs without errors on my machine.

kispaljr:
Also, after we merge this, we will have to adjust the go.mod files in the Nephio project. (But first, we have to merge this, so that a commit hash is assigned to the new single module)

PR #113, opened 2024-09-23
change trigger for porchctl release build

radoslawc:
It will only trigger on tag named in scheme vX.Y.Z


efiacor:
required for testing updates

PR #116, opened 2024-09-24
GitHub Actions exclude -pre tags

radoslawc:
This will exclude -pre tags from triggers in release GitHub Actions


liamfallon (commented):
/approve
/lgtm

PR #131, opened 2024-11-06
Tag container images built from git commits tagged with vX.Y.Z with the same vX.Y.Z tag

kispaljr:


radoslawc:
I've tested it on my porch fork and it works fine
/lgtm

PR #132, opened 2024-11-06
Automatically trigger the "release" Github action by pushing pre-release tags

kispaljr:


liamfallon (commented):
/approve
/lgtm

efiacor:
@radoslawc I think that this tag syntax is already supported. Can you confirm?

PR #142, opened 2024-11-24
Fix rendering big (>4MB) packages with function-runner

kispaljr:
Fixes https://github.com/nephio-project/nephio/issues/823

PR #121 successfully solved the issue of creating/updating/querying big packages via the porch API. However if a big package has a non-empty pipeline, the rendering of those large packages fail with an error message mentioning grpc max message size (see the issue referenced above). This PR fixes that.

(It also contains some minor refactoring that eliminates some unnecessarily long argument lists in function-runner.)

### DIFF
--- a/func/server/server.go
+++ b/func/server/server.go
@@ -68,7 +68,7 @@ func main() {
 	flag.StringVar(&o.pod.RegistryAuthSecretName, "registry-auth-secret-name", "auth-secret", "The name of the secret used for authenticating to custom registries")
 	flag.BoolVar(&o.pod.EnablePrivateRegistriesTls, "enable-private-registries-tls", false, "if enabled, will prioritize use of user provided TLS secret when accessing registries")
 	flag.StringVar(&o.pod.TlsSecretPath, "tls-secret-path", "/var/tmp/tls-secret/", "The path of the secret used in tls configuration")
-	flag.IntVar(&o.pod.MaxGrpcMessageSize, "max-request-body-size", 6*1024*1024, "Maximum size of grpc messages in bytes.")
+	flag.IntVar(&o.pod.MaxGrpcMessageSize, "max-request-body-size", 6*1024*1024, "Maximum size of grpc messages in bytes. Keep this in sync with porch-server's corresponding argument.")
 
 	flag.Parse()
 
--- a/func/server/server.go
+++ b/func/server/server.go
@@ -37,35 +37,49 @@ const (
 	wrapperServerImageEnv = "WRAPPER_SERVER_IMAGE"
 )
 
-var (
-	port                       = flag.Int("port", 9445, "The server port")
-	functions                  = flag.String("functions", "./functions", "Path to cached functions.")
-	config                     = flag.String("config", "./config.yaml", "Path to the config file.")
-	enablePrivateRegistries    = flag.Bool("enable-private-registries", false, "if true enables the use of private registries and their authentication")
-	registryAuthSecretPath     = flag.String("registry-auth-secret-path", "/var/tmp/config-secret/.dockerconfigjson", "The path of the secret used for authenticating to custom registries")
-	registryAuthSecretName     = flag.String("registry-auth-secret-name", "auth-secret", "The name of the secret used for authenticating to custom registries")
-	enablePrivateRegistriesTls = flag.Bool("enable-private-registries-tls", false, "if enabled, will prioritize use of user provided TLS secret when accessing registries")
-	tlsSecretPath              = flag.String("tls-secret-path", "/var/tmp/tls-secret/", "The path of the secret used in tls configuration")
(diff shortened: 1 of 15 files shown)


func/server/server.go
```diff
@@ -37,34 +37,49 @@ const (
...
+type options struct {
+	// The server port
+	port int
+	// The runtime(s) to disable. Multiple runtimes should separated by `,`.
+	disableRuntimes string
+
+	// Parameters of ExecEvaluator
+	exec internal.ExecutableEvaluatorOptions
+	// Parameters of PodEvaluator
+	pod internal.PodEvaluatorOptions
```
nagygergo, line 49: Pedantic, so your call if you fix it, but o.pod.wrapperServerImage below were a bit  confusing whtether this is actual pod info, or something else. Maybe podeval?

func/server/server.go
```diff
@@ -109,7 +124,10 @@ func run() error {
 	klog.Infof("Listening on %s", address)
 
 	// Start the gRPC server
-	server := grpc.NewServer()
+	server := grpc.NewServer(
+		grpc.MaxRecvMsgSize(o.pod.MaxGrpcMessageSize),
```
nagygergo, line 128: I'm not good at grpc, but is sending multi-MB messages a good thing to do?

pkg/cmd/server/start.go
```diff
@@ -58,7 +58,7 @@ type PorchServerOptions struct {
 	RepoSyncFrequency                time.Duration
 	UseGitCaBundle                   bool
 	DisableValidatingAdmissionPolicy bool
-	MaxRequestBodySize               int64
+	MaxRequestBodySize               int
```
nagygergo, line 61: Any reason this must be auto-selected at compile time?

liamfallon (commented):
/approve

test/e2e/e2e_test.go
```diff
@@ -2217,6 +2217,105 @@ func (t *PorchSuite) TestPodEvaluatorWithFailure(ctx context.Context) {
 	}
 }
 
+func (t *PorchSuite) TestPodEvaluatorWithLargeObjects(ctx context.Context) {
```
nagygergo, line 2220: Should this be merged with the TestLargePackageRevision test case

Catalin-Stratulat-Ericsson:
/approve
looks good to me.
Thank you for the cleanup code on the arguments.

func/server/server.go
kispaljr: I am fine with renaming it. Would "WrapperServerConatinerImageName" sound better? Should we also rename the WRAPPER_SERVER_IMAGE envvar where the actual name coming from? The later is a breaking change so probably not. Btw., there is a comment next to the field in the PodEvaluatorOptions struct (that IDEs will find and display) that tries to explain the semantics of the field.

func/server/server.go
kispaljr: I don't know if it is good practice or not, probably sending it in chunks via a grpc stream may be preferred where it makes sense. But, in this case we already have the architecture and the interfaces between components (porch-server <-> function-runner and function-runner <-> wrapper-server) that this fix doesn't want to change, and technically it works. Redesigning  the architecture is an option, but this PR only meant to be a hotfix for a bug.

pkg/cmd/server/start.go
kispaljr: The only reason is that GRPC mas-size expects an `int`, the apiserver runtime max-request-size expects an `int64` and I considered safer to convert from int to int64, than the other way around.

test/e2e/e2e_test.go
kispaljr: Well, I believe this new testcase already in its current form incorporates what is tested in `TestLargePackageRevision`, making the later redundant. (Not a surprise, since some of the code was copy-pasted from there.) I didn't want to delete any tests, but if others agree that I am right, then we can delete TestLargePackageRevision.

kispaljr: If we leave only one testcase, then we can rename the new test to `TestLargePackageRevision` if it explains the goal of the test better

nagygergo: Either way is fine with me.

kispaljr: I merged the two test cases: the new test case will always test updating a PRR with large amount of data, and skip only the Pod Evaluator test if tests are running without a function-runner

liamfallon:
/lgtm cancel
Waiting for the "draft" flag to be removed.

kispaljr:
Sorry @liamfallon for adding the draft flag last-minute. I made a last-minute test, but everything seems fine, so I removed the draft flag again

liamfallon:
> Sorry @liamfallon for adding the draft flag last-minute. I made a last-minute test, but everything seems fine, so I removed the draft flag again

Perfect.

PR #151, opened 2024-12-05
Pass default-image-prefix to function runners

mozesl-nokia:
The default-image-prefix server argument was not passed along to the KRM function runner previously. This PR is a fix for that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ __debug*
 ### VisualStudioCode Patch ###
 # Ignore all local history of files
 **/.history
+
+### Jetbrains IDEs ###
+.idea/*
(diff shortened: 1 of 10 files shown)


liamfallon (commented):
It seems very complex to achieve even simple things in Porch.

/approve

liamfallon:
@mozesl-nokia you will need to sign the CLA before we can merge this PR.

kispaljr:
Just a quick note: we are aware of this constant in the podevaluator: https://github.com/nephio-project/porch/blob/6d34cf80c7c6b80765c1c228035d507e72b1e064/func/internal/podevaluator.go#L65

I recommended not to touch that in this PR, but handle it in a future, unrelated PR. The rationale behind that is the following: 

in my understanding that constant is used to determine if special auth parameters and/or a special TLS ca.crt should be used for a registry or not. That means this constant shouldn't be set to the default registry passed as a command line argument, because it can be a private repository. Instead we should allow specifying which custom ca.crt should be used for which registry, but that is out-of-scope of this bugfix. @Catalin-Stratulat-Ericsson could you verify that?

PR #165, opened 2024-12-24
fix for #631 where slight differences seen in Kptfile upstream format…

mansoor17syed:
This push fixes issue #631


liamfallon:
Hi @mansoor17syed thanks for taking on this issue. The `kpt alpha repo` and `kpt alpha rpkg` commands have been deprecated by the `porchctl` command. I think this issue is likely to be still there in `porchctl`. I wonder if you could verify that. It's probably the `porchctl rpkg get` and porchctl rpkg clone` commands that should be checked.

efiacor (commented):
Hi @mansoor17syed . Changes look good. Do we have a unit test covering this?

mansoor17syed:
@efiacor 
yes added the UT cases, Please have a loot and let me know your feedback

efiacor:
> @efiacor @liamfallon yes added the UT cases, Please have a loot and let me know your feedback

Thanks for adding the test. Any chance we could just do a simple unit test instead of adding it to the e2e suite? I don't think we get sonar coverage from those.

mansoor17syed:
Hi @efiacor  and Team,

Can you Please have a look now worked on the review comments.

PR #167, opened 2025-01-09
fix for #736 Implement missing Modify event on the porch Repo Custom Resource 

mansoor17syed:


pkg/registry/porch/background.go
```diff
@@ -137,7 +137,7 @@ func (b *background) updateCache(ctx context.Context, event watch.EventType, rep
 		return b.cacheRepository(ctx, repository)
 	case watch.Modified:
 		klog.Infof("Repository modified: %s:%s", repository.ObjectMeta.Namespace, repository.ObjectMeta.Name)
-		// TODO: implement
+		return b.cacheRepository(ctx, repository)
```
liamfallon, line 140:
```suggestion
		var repoList configapi.RepositoryList
		if err := b.coreClient.List(ctx, &repoList); err != nil {
			return err
		}
		return b.cache.UpdateRepository(ctx, repository)
```

liamfallon: I think if we include a check on the core client to make sure that the repositories can still be listed (that is that the core client is still alive) then it would be good. See the code for the `delete` case, that is what they are doing there.

liamfallon (commented):
/approve

mansoor17syed:
Hi Team,
Could you please share your feedback or advise on the next steps to move forward?

liamfallon:
> Hi Team, Could you please share your feedback or advise on the next steps to move forward?

@mansoor17syed the unit tests are failing, I just reran them there now, it might be just an intermittent failure.

liamfallon:
The tests passed that time.

/approve
/lgtm

PR #168, opened 2025-01-10
Fixing typo in path pkg/cache/memory/cache.go at line 174

AshaRajput-27:
I see there was a typo error at line number 174 in file path pkg/cache/memory/cache.go
It should be CloseRepository instead of OpenRepository.


efiacor:
@AshaRajput-27 you will also need to sign the LF CLA to proceed

AshaRajput-27:
Hi @efiacor , I have done the CLE signing but not sure if it correct or why its not getting updated.

liamfallon (commented):
/approve
/lgtm

Thanks for the PR!

PR #174, opened 2025-01-23
Proposal- Porch Performance Testing Framework - Operation Flow with Prometheus

mansoor17syed:
**Proposal: Porch Performance Testing Framework with Prometheus Integration**

**Overview**
This proposal introduces a comprehensive performance testing framework for Porch, designed to measure and monitor repository and package management operations using Prometheus. The framework provides automated setup, detailed metrics collection, and extensive reporting capabilities.

Basic Test Execution
```
go test -v ./... \
  --repos=2 \
  --packages=5 \
  --packageRevisions=4
  
```
**Goals**

1. Establish automated performance testing for Porch operations
2. Implement comprehensive metrics collection using Prometheus
3. Provide detailed operation monitoring and reporting
4. Enable configurable test scenarios

**Implementation Details**

1. Framework Components

- Automated Setup
- Prometheus Configuration
- Automatic container creation and initialization
- Custom prometheus.yml configuration
- Access: http://localhost:9090
- Metrics Server
- Port: :2113
- Endpoint: http://localhost:2113/metrics
- Logging System
- Directory: ./logs
- Format: porch-metrics-YYYY-MM-DD-HH-MM-SS.log
- Console and file logging

**2. Monitored Operations**

**Repository Management**

- Create Gitea Repository
- Create Porch Repository
- Wait Repository Ready
- Delete Repository

**Package Management**

- Create PackageRevision
- Update to Proposed
- Update to Published
- Delete PackageRevision

**Metrics Collection**

- Performance Metrics
- Operation duration tracking
- Success/failure counting
- Resource utilization monitoring
- Error tracking and reporting

**Prometheus Integration**

```
scrape_configs:
  - job_name: 'porch_metrics'
    static_configs:
      - targets: ['host.docker.internal:2113']
    scrape_interval: 1s
```

**Benefits**

- Automated performance testing
- Comprehensive metrics collection
- Real-time monitoring capabilities
- Detailed reporting and analysis
- Configurable test scenarios

**Future Enhancements**

- Additional metric types
- Enhanced visualization options
- Automated performance regression detection
- Extended test scenario support

image
image
image
image
image
image
image
image

### DIFF
--- a/test/performance/README.md
+++ b/test/performance/README.md
@@ -1,0 +1,136 @@
+Copyright 2024 The Nephio Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+
+# Porch Performance Testing Setup Guide
+
+## Prerequisites
+- Docker
+- Kubernetes CLI (kubectl)
+- Go development environment
+- Access to GitHub repositories
+
+## 1. Set Up Development Environment
+Run the following script to set up the development environment:
+```bash
+https://github.com/nephio-project/porch/blob/main/scripts/setup-dev-env.sh
+```
+
+## 2. Build and Deploy Porch
+Execute the command below to build and deploy Porch in a Kubernetes Kind cluster:
+```bash
+make run-in-kind
+```
+Once deployment is complete, you should see the related pods running.
+
+## 3. Create a Demo Namespace
+Create a namespace for the demo setup:
+```bash
+kubectl create namespace porch-demo
+```
+
+## 4. Define Gitea Repositories in Porch
+Apply the repository configuration file:
+```bash
+kubectl apply -f examples/tutorials/starting-with-porch/porch-repositories.yaml
+```
+
+Note: The Gitea credentials secret is automatically created when running the tests. The configuration can be found in `gitea-secret.yaml`.
+
+## 5. Verify Repository Setup
+Check the status of repositories in the porch-demo namespace:
+```bash
+kubectl get repositories -n porch-demo
+```
+
+## 6. Run Performance Tests
+Navigate to the performance test directory:
+```bash
+cd test/performance/
+```
+
+Execute the performance test command:
+```bash
+E2E=1 go test -v ./... -repos=4 -packages=4 -timeout 60m
+```
+
+Parameters:
+- `repos`: Number of repositories to test
+- `packages`: Number of package revisions in each repository
+
+Once the tests begin, the logs can be observed in logs dir a sample is shared below.
+```
+=== Iteration 0 Results ===
+Operation                  Duration    Status
+--------------------------------------------------
+Create Gitea Repository   331ms      Success
+Create Porch Repository   7ms        Success
+Wait for Porch Repository Ready 2.008s     Success
+
+=== Iteration 1 Results ===
+Operation                  Duration    Status
+--------------------------------------------------
+Create PackageRevision    457ms      Success
+Update to Proposed        561ms      Success
+Update to Published       699ms      Success
+Delete PackageRevision    3.171s     Success
+
+=== Iteration 1 Results ===
+Operation                  Duration    Status
+--------------------------------------------------
+Delete Repository         5ms        Success
+
+=== Consolidated Performance Test Results ===
+Operation                  Min         Max         Avg         Total
+------------------------------------------------------------------------
+Update to Published       699ms       699ms       699ms       699ms
+Delete PackageRevision    3.171s      3.171s      3.171s      3.171s
+Create Gitea Repository   331ms       331ms       331ms       331ms
+Create Porch Repository   7ms         7ms         7ms         7ms
+Wait for Porch Repository Ready 2.008s      2.008s      2.008s      2.008s
+Create PackageRevision    457ms       457ms       457ms       457ms
+Update to Proposed        561ms       561ms       561ms       561ms
+```
+
+## 7. Verify Prometheus Targets
+Check the Prometheus targets:
+```
+http://localhost:9090/targets
+```
+You should be able to run PromQL queries here.
+
+## 8. Check Metrics Output
+Retrieve the captured metrics using:
+```bash
+curl -k http://localhost:2113/metrics
+```
+This should return all the collected performance metrics from the test execution.
+
+## Prometheus Setup
+
+The performance testing framework uses a Kubernetes-based Prometheus setup. The configuration is stored in `prometheus-manifests.yaml` and includes:
+
+- ConfigMap with Prometheus configuration
+- Deployment for the Prometheus pod
+- Service to expose Prometheus
+
+The Prometheus instance is automatically deployed in the `porch-demo` namespace when running the tests. You can access the Prometheus UI at:
+```
+http://localhost:9090
+```
+
+The configuration can be modified by editing `prometheus-manifests.yaml`. Key settings include:
+- Scrape interval: 1s
+- Resource limits: 512Mi memory, 200m CPU
+- NodePort: 30090
--- a/test/performance/metrics.go
+++ b/test/performance/metrics.go
@@ -1,0 +1,175 @@
+// Copyright 2024 The Nephio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package metrics
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/nephio-project/porch/api/generated/clientset/versioned"
+	porchapi "github.com/nephio-project/porch/api/porch/v1alpha1"
+	configapi "github.com/nephio-project/porch/api/porchconfig/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(porchapi.AddToScheme(scheme))
+	utilruntime.Must(configapi.AddToScheme(scheme))
+}
+
+// Helper functions for metrics collection and operations
+func getPorchClientset(cfg *rest.Config) (*versioned.Clientset, error) {
+	return versioned.NewForConfig(cfg)
+}
+
+func createGiteaRepo(repoName string) error {
+	giteaURL := "http://172.18.255.200:3000/api/v1/user/repos"
+	payload := map[string]interface{}{
+		"name":        repoName,
+		"description": "Test repository for Porch metrics",
+		"private":     false,
+		"auto_init":   true,
+	}
+
+	jsonPayload, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("failed to marshal payload: %v", err)
+	}
+
+	req, err := http.NewRequest("POST", giteaURL, bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %v", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.SetBasicAuth("nephio", "secret")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to create repo: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusCreated {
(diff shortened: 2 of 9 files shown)


mansoor17syed:
Hi team,
I've worked on the Porch Performance Testing Framework with Prometheus integration. It includes automated setup, metrics collection, and detailed logging. Please review the changes and share your feedback.

efiacor:
Hi @mansoor17syed .
Thanks for setting this up. Really nice to have.
Couple of points. 
Is the plan for this to be run as an e2e test? If so, we should prob align with how the others are doing it.
See [here ](https://github.com/nephio-project/porch/blob/main/.github/workflows/porch-e2e.yaml)for the main e2e gh action workflow.
So, it sets up the kind cluster, deploys porch with the Make target "run-in-kind", and then executes the tests.
They also have a flag to avoid these being executed as part of the unit test runs - "E2E=1 go test ..."

For interacting with gitea, we should avoid using the ip address.
https://github.com/nephio-project/porch/pull/174/files#diff-da6008a39400b7b8b7c9a7b68d4b0928faf93f44c3ca48ad1d29cb9cfb4ae40fR37

The existing e2e suite use a stubbed git deployment - https://github.com/nephio-project/porch/blob/main/test/e2e/suite.go#L556
Could we make use of this for your suite?

mansoor17syed:
Hi @efiacor 

Apologies for the delayed response—I was on vacation.

The main goal was to add metrics, but I wasn't sure how the community would receive them. So, I initially included them as part of the tests and proposed the changes.

I’m happy to modify or integrate these metrics directly into the code based on your feedback and requirements. Since you have a better perspective on this, I’d appreciate your guidance.

Looking forward to your response.

liamfallon:
Hi @mansoor17syed , the unit tests are failing on this one.

efiacor:
> Hi @efiacor
> 
> Apologies for the delayed response—I was on vacation.
> 
> The main goal was to add metrics, but I wasn't sure how the community would receive them. So, I initially included them as part of the tests and proposed the changes.
> 
> I’m happy to modify or integrate these metrics directly into the code based on your feedback and requirements. Since you have a better perspective on this, I’d appreciate your guidance.
> 
> Looking forward to your response.

Hi @mansoor17syed ,

No problem. So if we want to add this as a stand alone suite we will need some mech to setup the test cluster.
The e2e suites use a local kind cluster and a stubbed gitea. If you need to use a full gitea deployment then we have a pkg for that.
Also, to avoid these being run during the unit tests, we will need to add some mech to skip similar to here - https://github.com/nephio-project/porch/blob/main/test/e2e/e2e_test.go#L62

mansoor17syed:
1. Regarding cluster setup:

For running these performance tests, can I use the existing kind cluster setup from the e2e framework? 
I notice the e2e tests use a local kind cluster - should I integrate with that same mechanism?

2. Regarding Gitea:

The e2e suites use a stubbed gitea deployment. You mentioned there's a pkg for full gitea deployment - which approach would be more appropriate for these performance tests? Should I:
a) Use the existing stubbed gitea from e2e suite
b) Use the full gitea deployment pkg you mentioned

3. Regarding test organization:

Would you prefer these performance tests to:
a) Run as part of the existing e2e test suite
b) Remain as a standalone suite but use the e2e infrastructure
c) Have a completely separate setup

Looking forward to your thoughts!

efiacor:
Hi @mansoor17syed . Would be good to pick this up again for R5 if you are willing. Thanks

mansoor17syed:
Hi team,

There were some test failures, which I have now fixed. I have also added the README.md file too.

Please review and share your feedback.

efiacor:
Hi @mansoor17syed 
Could you do a rebase and maybe squash the commits to avoid the repetition of "Proposal- Porch Performance Testing Framework - Operation Flow with Prometheus" message. Thanks

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+
+	// Wait for package to initialize
+	time.Sleep(5 * time.Second)
+	debugPackageStatus(t, c, ctx, namespace, newPkg.Name)
+
+	// Propose the package
+	start = time.Now()
+	var pkg porchapi.PackageRevision
+	err = c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: newPkg.Name}, &pkg)
```
efiacor, line 146: Should we include the gets in the time calc?
Same for other gets below.

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+	metrics = append(metrics, OperationMetrics{
+		Operation: "Delete PackageRevision",
+		Duration:  time.Duration(duration * float64(time.Second)),
+		Error:     err,
+	})
+
+	return metrics
+}
+
+func setupMonitoring(t *testing.T) error {
```
efiacor, line 208: Would it not be better to deploy a basic in memory k8s prometheus pod, etc?
We should have these configs somewhere we can tweak them if needed.

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+		return fmt.Errorf("failed to create prometheus.yml: %v", err)
+	}
+
+	// Execute Docker commands
+	cmds := []struct {
+		name string
+		cmd  string
+		args []string
+	}{
+		{"network create", "docker", []string{"network", "create", "monitoring"}},
```
efiacor, line 231: Not a fan of this direct docker calls.

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+	start = time.Now()
+	repo := &configapi.Repository{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      repoName,
+			Namespace: namespace,
+		},
+		Spec: configapi.RepositorySpec{
+			Type: "git",
+			Git: &configapi.GitRepository{
+				Repo:   fmt.Sprintf("http://172.18.255.200:3000/nephio/%s", repoName),
```
efiacor, line 59: I think the URL should be configurable

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+	numRepos    = flag.Int("repos", 1, "Number of repositories to create")
+	numPackages = flag.Int("packages", 5, "Number of packages per repository")
+)
+
+func createAndSetupRepo(t *testing.T, ctx context.Context, c client.Client, namespace, repoName string) []OperationMetrics {
+	var metrics []OperationMetrics
+	start := time.Now()
+
+	// Create Gitea repo
+	err := createGiteaRepo(repoName)
```
efiacor, line 35: Do we care how long it takes gitea to create the repo?

test/performance/metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusNoContent && resp.StatusCode != http.StatusNotFound {
+		return fmt.Errorf("failed to delete repo, status: %d", resp.StatusCode)
+	}
+
+	return nil
+}
+func waitForRepository(ctx context.Context, c client.Client, t *testing.T, namespace, name string, timeout time.Duration) error {
```
efiacor, line 129: *waitForPorchRepository

test/performance/metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	utilruntime.Must(configapi.AddToScheme(scheme))
+}
+
+// Helper functions for metrics collection and operations
+func getPorchClientset(cfg *rest.Config) (*versioned.Clientset, error) {
+	return versioned.NewForConfig(cfg)
+}
+
+func createGiteaRepo(repoName string) error {
+	giteaURL := "http://172.18.255.200:3000/api/v1/user/repos"
```
efiacor, line 38: Again, URL should be dynamic. Alternatively, we use a gitea client - 
https://pkg.go.dev/code.gitea.io/sdk/gitea#Client.CreateRepo

test/performance/README.md
```diff
@@ -0,0 +1,80 @@
...
+Create a secret to store Gitea credentials within the demo namespace:
+```bash
+kubectl create secret generic gitea \
+    --namespace=porch-demo \
+    --type=kubernetes.io/basic-auth \
+    --from-literal=username=nephio \
+    --from-literal=password=secret
+```
+
+## 5. Define Gitea Repositories in Porch
```
efiacor, line 38: Actually, I don't think this is necessary

test/performance/README.md
```diff
@@ -0,0 +1,80 @@
...
+```bash
+cd porch/test/performance/
+```
+
+Execute the performance test command:
+```bash
+E2E=1 go test -v ./... -repos=4 -packages=4 -timeout 60m
+```
+
+Parameters:
```
efiacor, line 62: Indicate all possible inputs and default values.

test/performance/README.md
```diff
@@ -0,0 +1,80 @@
...
+- `repos`: Number of repositories to test
+- `packages`: Number of package revisions in each repository
+
+Once the tests begin, the metrics can be observed.
```
efiacor, line 66: Give an example of the output and details of the log files created.

test/performance/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+http://localhost:9090/targets
+```
+You should be able to run PromQL queries here.
+
+## 9. Check Metrics Output
+Retrieve the captured metrics using:
+```bash
+curl -k http://localhost:2113/metrics
+```
```
efiacor, line 79: Add info regarding the queries file and how to use them

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      repoName,
+			Namespace: namespace,
+		},
+		Spec: configapi.RepositorySpec{
+			Type: "git",
+			Git: &configapi.GitRepository{
+				Repo:   fmt.Sprintf("http://172.18.255.200:3000/nephio/%s", repoName),
+				Branch: "main",
+				SecretRef: configapi.SecretRef{
```
efiacor, line 61: We could define this secret locally and create it as part of the test setup.

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,518 @@
...
+		Duration:  time.Duration(duration * float64(time.Second)),
+		Error:     err,
+	})
+
+	if err == nil {
+		repositoryCounter.Inc()
+		start = time.Now()
+		err = waitForRepository(ctx, c, t, namespace, repoName, 60*time.Second)
+		duration = time.Since(start).Seconds()
+		recordMetric("Wait Repository Ready", repoName, "", duration, err)
```
efiacor, line 83: ... Porch Repository Ready ...

test/performance/README.md
```diff
@@ -0,0 +1,80 @@
...
+    --namespace=porch-demo \
+    --type=kubernetes.io/basic-auth \
+    --from-literal=username=nephio \
+    --from-literal=password=secret
+```
+
+## 5. Define Gitea Repositories in Porch
+Apply the repository configuration file:
+```bash
+kubectl apply -f porch-repositories.yaml
```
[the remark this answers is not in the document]
mansoor17syed: Corrected

test/performance/README.md
```diff
@@ -0,0 +1,80 @@
...
+Check the status of repositories in the porch-demo namespace:
+```bash
+kubectl get repositories -n porch-demo
+```
+
+
+## 7. Run Performance Tests
+Navigate to the performance test directory:
+```bash
+cd porch/test/performance/
```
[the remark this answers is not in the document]
mansoor17syed: Corrected

test/performance/README.md
mansoor17syed: There are no default params and these are the only two possible params for now to run tests

test/performance/README.md
mansoor17syed: Done

test/performance/README.md
mansoor17syed: yes there is a query file and data/info about promql attached to the git PR

test/performance/porch_metrics_test.go
mansoor17syed: Done

test/performance/porch_metrics_test.go
efiacor: But technically the "Get" is not part of the package proposal right?

test/performance/README.md
efiacor: I think there are default values - https://github.com/nephio-project/porch/pull/174/files#diff-2108193b521a1a0005d0859c990a880d78a8b44b726992a80c7b69784842b11eR38

test/performance/porch_metrics_test.go
mansoor17syed: You are pointing out that we're including the Get operation in our timing calculation for the "Update to Proposed" metric. This is technically incorrect because:
The Get operation is just retrieving the current state of the package
The actual "proposal" operation is the Update that changes the lifecycle state
Including the Get in the timing makes the metric less accurate for measuring the actual proposal operation

The correct approach would be to:
First get the package (outside the timing)
Then start timing only the actual update operation

efiacor: Apologies. My first comment was unclear. I was pointing out that we prob shouldn't include the Get operation in the timings, like you mention above.

mansoor17syed: No worries will fix it

test/performance/porch_metrics_test.go
efiacor: Not sure what your reply refers to. We can define a secret in a yaml file in the same dir and creat it as part of the test setup.

test/performance/porch_metrics_test.go
```diff
@@ -0,0 +1,532 @@
...
+					time.Sleep(5 * time.Second)
+					debugPackageStatus(t, c, ctx, namespace, pkg.Name)
+				}
+			}
+		}
+	}
+
+	// Delete package
+	start = time.Now()
+	err = c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: pkg.Name}, &pkg)
```
efiacor, line 202: All of the "Get"s should be outside the timers.

test/performance/porch_metrics_test.go
mansoor17syed: Yes, you're right—this is a good solution. and this is exactly what I have in mind.

To implement this change, we would need to:

Create a Prometheus ConfigMap

Create a Prometheus Deployment

Create a Service to expose Prometheus

Use the Kubernetes client to manage these resources

 However, the goal of this PR is to initially implement metrics and capture them as part of test or performance test. If this approach works well, the next step would be to integrate it into the code so that each operation generates metrics. These metrics would then be collected using an external Prometheus instance. This aligns with what the service assurance team is aiming to achieve.

test/performance/porch_metrics_test.go
mansoor17syed: Done

PR #175, opened 2025-01-24
Prevent creating "main" tag and release on Github

kispaljr:
A tag named "main" (same as the branch name "main") in the GitHub repo confuses git clients and scripts executing git operations, so we should avoid it if possible.

This 3 character long PR tries to fix the GH action that regularly recreates that tag,

### DIFF
--- a/.github/workflows/porchctl-dev-release.yaml
+++ b/.github/workflows/porchctl-dev-release.yaml
@@ -19,6 +19,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/porchctl-dev-release.yaml
```diff
@@ -45,9 +46,8 @@ jobs:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: porchctl.tgz
           asset_name: porchctl.tgz
-          tag: ${{ github.ref }}
+          tag: dev
```
efiacor, line 49: Would we need to align test-infra to pick up the correct bin?
https://github.com/nephio-project/test-infra/blob/main/e2e/provision/playbooks/cluster.yml#L141

kispaljr: Yes, definetly. Will send a co-PR.

liamfallon (commented):
/approve

PR #177, opened 2025-02-06
Use full resource paths in porchctl repo/rpkg get

mozesl-nokia:
Encountered an issue where 2 resources had the name "repository" in a cluster and `porchctl repo get` consistently returned the wrong ones. This PR is to solve that issue by appending the group to the query (for both repository and packagerevision).


liamfallon (commented):
/approve

PR #178, opened 2025-02-06
Add additional info to error when blockOwnerDeletion is enabled

mozesl-nokia:
Related issue: [#835](https://github.com/nephio-project/nephio/issues/835)
This change is meant more as a mitigation than a solution to the problem, returning a more clear error message for a PackageRevision create/update failure when blockOwnerDeletion is enabled.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -198,7 +198,7 @@ func (cad *cadEngine) CreatePackageRevision(ctx context.Context, repositoryObj *
 	}
 	pkgRevMeta, err = cad.metadataStore.Create(ctx, pkgRevMeta, repositoryObj.Name, repoPkgRev.UID())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/engine/engine.go
```diff
@@ -218,6 +221,15 @@ func ensureUniqueWorkspaceName(obj *api.PackageRevision, existingRevs []reposito
 	return nil
 }
 
+func anyBlockOwnerDeletionSet(pr *api.PackageRevision) bool {
```
liamfallon, line 224: Could I suggest that you move this function into the go file https://github.com/nephio-project/porch/blob/main/pkg/repository/util.go?

This function isn't specific to the engine and could be used in other parts of the code.

mozesl-nokia: Done

liamfallon (commented):
/approve

PR #179, opened 2025-02-10
Removed upstream, mutators and validators from Repository API specification

gkumar78:
Nephio docs repo is also being updated. Found mutators for Repository mentioned at various places in Porch documentation


gkumar78:
Fixes https://github.com/nephio-project/nephio/issues/830

liamfallon (commented):
/approve

PR #182, opened 2025-02-14
changed mutex in cache repository to read write mutex so that all concurrent reads are allowed

sriharshabm:
As of now normal mutex is used in cache repository which will make multiple read to wait for mutex. Changed mutex in cache repository to read write mutex so that all concurrent reads are allowed.

Also. Added missing mutex unlock in case of failure

### DIFF
--- a/pkg/cache/memory/repository.go
+++ b/pkg/cache/memory/repository.go
@@ -54,7 +54,7 @@ type cachedRepository struct {
 
 	lastVersion string
 
-	mutex                  sync.Mutex
+	mutex                  sync.RWMutex
 	cachedPackageRevisions map[repository.PackageRevisionKey]*cachedPackageRevision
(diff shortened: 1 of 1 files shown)


mansoor17syed:
This is a good optimization because:

1. Read operations can now happen concurrently instead of being serialized
2. Only write operations will block other operations
3. Fixes a potential mutex leak in error paths

However, let me suggest some review points:

1. Verify Lock Usage
2. Ensure all read operations use RLock()/RUnlock()
3. Ensure all write operations use Lock()/Unlock()
4. Double check all error paths have proper unlocking


Test Coverage (Could be a separate PR)
1. Add concurrent read tests to verify multiple reads work simultaneously
2. Add mixed read/write tests to verify proper locking behavior
3. Add stress tests to catch any race conditions if possible

liamfallon (commented):
/approve

It would be good to get this into R4

PR #193, opened 2025-03-07
Make commit author and email customisable for git repositories

mozesl-nokia:
This PR aims to make the commits that porch makes to git repositories more customisable. Previously, the author and email fields were both populated by the authenticated kubernetes user's name (which will still be the case if no author or email is set in the repository object's spec).
This change should make it much easier to create test data like those in `pkg/externalrepo/git/testdata`.
I also bumped the api-related generated files, which in turn revealed that some shebangs broke from #189, so I fixed those as well.


pkg/externalrepo/git/git.go
```diff
@@ -162,6 +173,26 @@ func OpenRepository(ctx context.Context, name, namespace string, spec *configapi
...
+// makeUserInfoProvider creates a UserInfoProvider based on the provided GitRepository spec.
+// If the spec contains no author or email, def will be returned instead.
+// If only one of author or email is present, the one present will be used in the other's place.
+func makeUserInfoProvider(spec *configapi.GitRepository, def repository.UserInfoProvider) repository.UserInfoProvider {
+	if spec.Author != "" || spec.Email != "" {
+		ui := &gitUserInfoProvider{
+			Author: spec.Author,
+			Email:  spec.Email,
+		}
+		if ui.Author == "" {
```
liamfallon, line 185: Just wondering why you need lines 185 to 192. Wouldn't it be better to leave the fields blank if they're not specified, partucularly setting the email to be the author as it's very likely to be an invalid email.

nagygergo:
This means that all commits to a repository will be authored by the same user, right? 
I feel that there are some underlying assumptions that are not assured by e2e test cases. 
Things like multiple porch repositories in the same namespace pointing at the same git repositories, or porch repositories in multiple namespaces pointing at the same git repository. Probably those behaviors would be good to secure with some test cases, otherwise there might be a drift here.

pkg/externalrepo/git/git.go
mozesl-nokia: It's more or less for parity with the original, which uses the user name for the email field as well.
https://github.com/nephio-project/porch/blob/155acc438de7c7dfc603d17d2f746d273330cdb3/pkg/registry/porch/userinfo.go#L42-L45
But leaving it blank or having a hardcoded default would be better imo.

mozesl-nokia:
@nagygergo
> This means that all commits to a repository will be authored by the same user, right?

Yes, commits going through the same _porch_ repository.

> I feel that there are some underlying assumptions that are not assured by e2e test cases. Things like multiple porch repositories in the same namespace pointing at the same git repositories, or porch repositories in multiple namespaces pointing at the same git repository. Probably those behaviors would be good to secure with some test cases, otherwise there might be a drift here.

I'm not entire sure about your concern (although, writing some tests would be a good idea). In your cases the difference would be that the same porch instance could be making commits to the same git repo under multiple authors/emails. Could this cause issues somewhere?

liamfallon:
@mozesl-nokia Sorry for not getting back sooner, we had some holidays here.

I think it would be best to do as follows:
- When the email is specified and the name is not, then the email can be copied to the name field
- When the name is specified and the email is not, its probably best to leave the email field blank

liamfallon:
Looks great! I'm happy to approve once Sonar passes.

mozesl-nokia:
`TestPublishedLatest` seems to be flaky and it looks like I don't have permission to retest.

PR #202, opened 2025-03-12
Handle invalid kpt annotations more gracefully, handle tags with messages

mozesl-nokia:
These two issues were discovered when someone was trying to import git repositories into porch that have either been "made porch compliant" by hand, or by some controller/operator other than porch.

The first issue was that if there was one commit in a git repository that had an unparsable kpt annotation (e.g. bad json) the package discovery would break for the entire repository.

In this PR I've made it so that porch is much more lenient when it comes to invalid annotations, such as allowing empty task lists and defaulting the workspace name to the revision name even when there was an error walking the commits. I have also removed the type alias for WorkspaceName, as it seemed redundant and it was just being cast back and forth everywhere.

Another related issue was discovered where manually creating a tag with a message broke discovery. This is also fixed in this PR.

Other smaller changes:
- Introduced an "error collector" which helps collect and chain together errors that are "on the same level", and should not wrap each other.
- Wherever I modified the code, I changed the returned errors to ones made with https://github.com/pkg/errors. This gives us stack traces if printed with `%+v`. I only kept this in one place currently (pkg/registry/porch/packagecommon.go:89), but it can be expanded upon depending on the verbosity we want to log.


liamfallon:
Hi @mozesl-nokia , this looks great from a desk check. Let me download the PR and run through it locally.

efiacor:
Hi @mozesl-nokia , looks good. Could you please push a minor change to trigger the sonar gh action?

PR #207, opened 2025-03-13
Port main E2E test suite to testify

mozesl-nokia:
The main E2E test suite was a wrapper around `testing.T` and mostly mimicking testify's functionality. In this PR I "ported" the suite over to testify, making it more IDE friendly in my experience.

I also removed a lot of redundant context passing, pushing it down to the helper functions instead. Furthermore, I fixed the lint issues in the tests and removed the exclusion in the github action.

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -51,4 +51,4 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: v1.62.2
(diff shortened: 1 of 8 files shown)


.github/workflows/golangci-lint.yaml
```diff
@@ -51,4 +51,4 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: v1.62.2
-          args: --timeout=10m --go=${{ env.GO_VERSION }} --exclude-dirs=test --exclude-generated=true
```
efiacor, line 54: Not sure this is necessary really.

PR #212, opened 2025-03-17
Internalize Kpt/KRM function go SDK from GoogleContainerTools

kispaljr:
This go library allows us to manipulate KRM manifests without loosing formatting, like comments, indentation or the order of fields. 
* It is heavily used in Nephio, and internalizing it here would eliminate the need to trying to add improvements to the library via an external overlay, in other words most of https://github.com/nephio-project/nephio/tree/main/krm-functions/lib can be integrated into the newly internalized lib
* right now I need it to implement Kptfile readiness gate and condition handling in https://github.com/nephio-project/porch/pull/194, 
* the library is heavily used in my new PackageVariant implementation, and this PR can be considered the first in a train of PRs that leads to proposing that controller to porch

The original package is effectively unmaintained for years by now.


efiacor:
@kispaljr are we sure about this one? Is there no chance we can engage the GoogleContainerTools guys on this code and keep it as an external dependency?

kispaljr:
> @kispaljr are we sure about this one? Is there no chance we can engage the GoogleContainerTools guys on this code and keep it as an external dependency?

There is no reply from them for more than a year by now even for the question of when will this be migrated to the kptdev GH organization. No new commits, nobody responds to PRs.

kispaljr:
We also need this to be able to update dependencies (same reason why the built-in KRM function code was vendored in)

kispaljr:
And we cannot even send a PR there, unless we sign the Google CLA, that is forbidden e.g. for me.

kispaljr:
> I wonder could the GoogleContainerTools things we need be moved to the https://github.com/kptdev project?

Yes, absolutely. I asked about the ETA of the migration in Oct 2023, and still no reply: https://github.com/kptdev/kpt/issues/4034
Unfortunately there aren't any active maintainers currently of the kptdev organization. If this changes we can move the SDK to that org.

sonar-project.properties
```diff
@@ -9,7 +9,7 @@ sonar.language=go
 sonar.sources=pkg, func, controllers, internal, cmd, build, deployments/porch, api
 
 # Exclude files if needed
-sonar.exclusions=**/test/*, **/examples/*, **/scripts/*, **/third_party/*, **/*_test.go, **/testing*, **/generated/**, **/testdata/**, **/*zz_generated.*
+sonar.exclusions=**/test/*, **/examples/*, **/scripts/*, **/third_party/GoogleContainerTools/kpt-functions-catalog/*, **/*_test.go, **/testing*, **/generated/**, **/testdata/**, **/*zz_generated.*
```
efiacor, line 12: sonar.exclusions=**/test/*, **/examples/*, **/scripts/*, **/*_test.go, **/testing*, **/generated/**, **/testdata/**, **/*zz_generated.*

lapentad: **/third_party/** 
If we add an extra '*' it should pick up any subdirectories, but I am not sure if it's too destructive

kispaljr:
@efiacor @liamfallon 
Here is my latest argument for this PR:
- current situation: we have an unmaintained dependency that we are using, but have to live with its outdated dependencies. We also cannot add improvements needed by Nephio to the code.
- if we fork it inside porch, but don't really maintain it after: 
  - the version of the lib in this PR is already a major improvement from Nephio's PoV compared to the original code
  - the go.mod of the vendored SDK is the same as porch's, so just by keeping porch's dependencies up-to-date, we implicitly also continuously improve/maintain the SDK without any extra effort
  - the vendored lib is only used by Nephio, so we only have to modify the code if it is needed by Nephio, something that we wouldn't be able to do at all, if we are not vendoring the lib

efiacor:
/approve

We can address the sonar check later. Not sure why it's not picking up the /third_party dir.

mozesl-nokia:
> the go.mod of the vendored SDK is the same as porch's, so just by keeping porch's dependencies up-to-date, we implicitly also continuously improve/maintain the SDK without any extra effort

This seems to cause kpt functions using older versions of kyaml (like apply-setters using v0.10.21) to break because of api incompatibility.

PR #238, opened 2025-05-06
Allow E2E tests to run behind an HTTP proxy

rendre-greyling:
The goal of this PR is to allow the E2E tests to be run behind a HTTP proxy and use internal mirrors of the repositories and registries.

### Additions:

- Repositories and registries are now specified by environment variables stored in an uncommitted .env file - see .env.template
- If no environment variable is specified, defaults are used.
- Addition of cloning package revision with a secret.
- Specifying of the Default Image Prefix (gcr.io prefix) from the .env file in the Makefile.
- Modification of how the Aliases are mapped in the builtin runtime to allow for custom image prefix.

### Fixes:

- Fixed a bug where when cloning a package revision with a secret, the namespace is not set.

These modifications should have no effect on E2E tests when no environment variables set. If environment variables are set - specifically the "PORCH_GCR_PREFIX_URL" - make run-in-kind should be executed.


rendre-greyling:
In addition, a proposed change which will decrease the number of environment variables and simplify the code: [Issue 907](https://github.com/nephio-project/nephio/issues/907)

efiacor:
https://github.com/nephio-project/porch/pull/240

liamfallon (commented):
/approve

Makefile
Catalin-Stratulat-Ericsson, line 1: I would recommend updating the Copyright all the files to match the latest year e.g. in this files case 2022-2025 instead of 2024.

pkg/engine/builtinruntime_test.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 func TestBuiltinRuntime(t *testing.T) {
-	br := newBuiltinRuntime()
+	br := newBuiltinRuntime("")
```
Catalin-Stratulat-Ericsson, line 29: why are we passing in empty strings here and in multiple areas instead of a variable of an empty string? is it more efficient one way or the other? looks a bit strange to me

Catalin-Stratulat-Ericsson:
Alot of new environment variables added for this new behind proxy E2E tests but there is no setup guide or documentation of how to set this up. or maybe im just being pedantic. would some documentation of how to utilize this be useful to accompany this PR? could anyone use this? how could they set this up for themselfs?

Catalin-Stratulat-Ericsson:
> Alot of new environment variables added for this new behind proxy E2E tests but there is no setup guide or documentation of how to set this up. or maybe im just being pedantic. would some documentation of how to utilize this be useful to accompany this PR? could anyone use this? how could they set this up for themselves?

Even just a small how-to or a configuration example file would go far in this regard :)

Catalin-Stratulat-Ericsson:
> > Alot of new environment variables added for this new behind proxy E2E tests but there is no setup guide or documentation of how to set this up. or maybe im just being pedantic. would some documentation of how to utilize this be useful to accompany this PR? could anyone use this? how could they set this up for themselfs?
> 
> Any objection to adding guide to the .env.template file and pointing to it with a comment in the suite_utils.go where all the constant for the environment variables are declared?

id say if its a guide on setting up E2E tests behind the proxy it belongs in the docs somewhere like here, https://docs.nephio.org/docs/porch/user-guides/ in a (setting up E2E tests behind proxy) section.

but if we are simply talking about a example inputs to the variables etc then id say that would suffice

rendre-greyling:
Added fixes and extra information in the .env.template file. Let me know if this is extra info is enough, or if a guide in the docs is needed. If it is, will work on that when I have extra time.

Catalin-Stratulat-Ericsson:
/approve
Thanks for the changes.
Looks good its a good bit more understandable what does what and how to configure up this now.

PR #242, opened 2025-05-07
Evict failed pods from cache and trigger fresh pod creation.

Bhargav-Singh:
### Problem

In the current implementation, function pods that enter a `Failed` state are not evicted from the cache. Consequently, Porch continues to reuse these failed pods for subsequent evaluations, leading to persistent failures even after the underlying issue (e.g., registry unavailability) has been resolved.

### Solution

This PR introduces the following changes:

* Detects and deletes `Failed` pods during evaluation request handling.
* Deletes and evicts `Failed` pods during garbage collection.
* Adds asynchronous pod deletion (`deleteFailedPod`) with cache safety checks.
* Ensures fresh pod creation is triggered immediately after eviction.
* Adds a unit test case under `TestPodManager` to validate `Failed` pod handling.

### DIFF
--- a/func/internal/podevaluator.go
+++ b/func/internal/podevaluator.go
@@ -311,7 +311,12 @@ func (pcm *podCacheManager) podCacheManager() {
 				err := pcm.podManager.kubeClient.Get(context.Background(), podAndCl.pod, pod)
 				deleteCacheEntry := false
 				if err == nil {
-					if pod.DeletionTimestamp == nil && net.JoinHostPort(pod.Status.PodIP, defaultWrapperServerPort) == podAndCl.grpcClient.Target() {
+					// If the cached pod is in Failed state, delete it and evict from cache to trigger a fresh pod.
+					if pod.Status.Phase == corev1.PodFailed {
+						klog.Warningf("pod %v/%v is in Failed state. Deleting and recreating", pod.Namespace, pod.Name)
+						go pcm.deleteFailedPod(pod, req.image)
+						deleteCacheEntry = true
+					} else if pod.DeletionTimestamp == nil && net.JoinHostPort(pod.Status.PodIP, defaultWrapperServerPort) == podAndCl.grpcClient.Target() {
 						klog.Infof("reusing the connection to pod %v/%v to evaluate %v", pod.Namespace, pod.Name, req.image)
 						req.grpcClientCh <- &clientConnAndError{grpcClient: podAndCl.grpcClient}
 						go patchPodWithUnixTimeAnnotation(pcm.podManager.kubeClient, podAndCl.pod, pcm.podTTL)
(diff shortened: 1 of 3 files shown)


liamfallon (commented):
Thanks for authoring this @Bhargav-Singh. It looks good. I wonder if we could get some code coverage on it?

Bhargav-Singh:
> Thanks for authoring this @Bhargav-Singh. It looks good. I wonder if we could get some code coverage on it?

Thanks! I’ve added a test (`Failed_pod_is_deleted_and_new_one_is_created`) that exercises the new logic, and it passes. But strangely, the coverage report (both locally and in SonarQube) only shows type definitions as covered, not the actual function code like `podCacheManager()`.

Can I get some help figuring out why is it.

efiacor:
> > Thanks for authoring this @Bhargav-Singh. It looks good. I wonder if we could get some code coverage on it?
> 
> Thanks! I’ve added a test (`Failed_pod_is_deleted_and_new_one_is_created`) that exercises the new logic, and it passes. But strangely, the coverage report (both locally and in SonarQube) only shows type definitions as covered, not the actual function code like `podCacheManager()`.
> 
> Can I get some help figuring out why is it.

The unit tests are using fakes so won't provide any coverage metrics. There are some podeval e2e tests. They won't provide coverage either but may be more useful in terms actually of testing the change.

Bhargav-Singh:
> > > Thanks for authoring this @Bhargav-Singh. It looks good. I wonder if we could get some code coverage on it?
> > 
> > 
> > Thanks! I’ve added a test (`Failed_pod_is_deleted_and_new_one_is_created`) that exercises the new logic, and it passes. But strangely, the coverage report (both locally and in SonarQube) only shows type definitions as covered, not the actual function code like `podCacheManager()`.
> > Can I get some help figuring out why is it.
> 
> The unit tests are using fakes so won't provide any coverage metrics. There are some podeval e2e tests. They won't provide coverage either but may be more useful in terms actually of testing the change.

I've added e2e test for this. As expected still not getting the code coverage. Is this fine or something more should be done?

func/internal/podevaluator.go
```diff
@@ -419,6 +435,19 @@ func (pcm *podCacheManager) garbageCollector() {
 	}
 }
 
+// deleteFailedPod deletes a failed pod and removes it from cache if it is still the current pod.
+func (pcm *podCacheManager) deleteFailedPod(pod *corev1.Pod, image string) {
```
efiacor, line 439: Could we move as much as possible to helper functions and test those in isolation?
We have some mocks defined for external interfaces using mockery - https://github.com/nephio-project/porch/tree/main/test/mockery/mocks/external/sigs.k8s.io/controller-runtime/pkg/client

We could add a mock for the grpc client here also.

PR #254, opened 2025-05-20
fix for #915 error out when --namespace is specified without a value

mansoor17syed:
1. Namespace Filtering Bug https://github.com/nephio-project/nephio/issues/915

Problem:
When listing or watching PackageRevision resources, results from other namespaces could appear, especially when using kubectl get packagerevisions -n <namespace> or -A.

2. Root Cause:
The code did not always properly filter PackageRevision objects by namespace, especially in the watch and list logic.

3. Testing
New Unit Tests:
Added tests in packagecommon_test.go to verify:
The watcher only forwards events for the correct namespace.
The list logic only returns revisions from the requested namespace.

### DIFF
--- a/pkg/registry/porch/packagecommon.go
+++ b/pkg/registry/porch/packagecommon.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/klog/v2"
@@ -95,6 +96,10 @@ func (r *packageCommon) listPackageRevisions(ctx context.Context, filter package
 				return err
 			}
 
+			if ns != "" && apiPkgRev.Namespace != ns {
+				continue
+			}
+
 			if selector != nil && !selector.Matches(labels.Set(apiPkgRev.Labels)) {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


pkg/registry/porch/packagecommon.go
```diff
@@ -95,6 +96,10 @@ func (r *packageCommon) listPackageRevisions(ctx context.Context, filter package
 				return err
 			}
 
+			if ns != "" && apiPkgRev.Namespace != ns {
```
kispaljr, line 99: I believe this should never happen.
 If ns != "", then `client.InNamespace(ns)` was already added to the list repositories API call above, so only Repositories in this namespace are listed, therefore only PackageRevisions from this namespace can be in the `revisions` list

pkg/registry/porch/packagecommon.go
```diff
@@ -144,8 +149,31 @@ func (r *packageCommon) listPackages(ctx context.Context, filter packageFilter,
 	return nil
 }
 
+// namespaceFilteringWatcher wraps an ObjectWatcher and filters events by namespace
+// Only events for the specified namespace are passed to the delegate watcher
```
kispaljr, line 153: I don't see how is this related to issue #915 . The problem doesn't seem that PackageRevisions from other namespaces appear in the watch stream that this PR seems to solve. 

Were objects from the wrong namespace observed in the watchstream?
If yes, please create a separate issue.

mansoor17syed: Thank you for your feedback and for pointing this out.
You are correct—the watch stream namespace filtering is not directly related to the original issue (#915).
We were under the impression that both kubectl and porchctl exhibited the same behavior regarding namespace filtering, which is why both fixes were included together.
Based on your feedback, we have now created a separate issue to track the watch stream filtering problem and moved the relevant code to a new PR.
This PR now only addresses the original namespace filtering bug for listing PackageRevision resources.
Thank you again for your guidance and clarification!

PR #256, opened 2025-05-22
Implement e2e tests for 3rd party updates

dgyorgy-nokia:
Updating 3rd party dependencies may break existing kpt functions because of api incompatibility. There should be e2e tests which go though all the used base kpt functions and tests them to see if they still work.

### DIFF
--- a/test/e2e/kpt_function_test.go
+++ b/test/e2e/kpt_function_test.go
@@ -1,0 +1,769 @@
+// Copyright 2025 The kpt and Nephio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


liamfallon (commented):
/approve

efiacor:
@dgyorgy-nokia looks like most of the new tests are failing. Can you run them successfully on your local setup?

dgyorgy-nokia:
> @dgyorgy-nokia looks like most of the new tests are failing. Can you run them successfully on your local setup?

Yes, it works locally, but I noticed that I left hardcoded the reference value for the used gitpackage, and probably that's why they fail. I uploaded a fix to use the configured reference value (PORCH_POD_EVAL_REF) from .env file.

PR #260, opened 2025-05-30
Resolve issue #915 where porchctl did not error if the --namespace flag was given without a value

mansoor17syed:
The issue 915 was caused because porchctl did not return an error when the --namespace flag was used without specifying a value (e.g., --namespace or --namespace=). The fix adds validation in the function of the rpkg get command—and other similar commands if needed—to check if the --namespace flag is present but empty. If the flag is detected with no value, the command now provides a clear error message and stops execution.

### DIFF
--- a/pkg/cli/commands/rpkg/get/command_test.go
+++ b/pkg/cli/commands/rpkg/get/command_test.go
@@ -19,7 +19,7 @@ import (
 	"testing"
 
 	porchapi "github.com/nephio-project/porch/api/porch/v1alpha1"
-	"github.com/nephio-project/porch/pkg/cli/commands/repo/get"
+	// repoGet "github.com/nephio-project/porch/pkg/cli/commands/repo/get" // Remove or comment out unless needed elsewhere
 
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
@@ -45,7 +45,7 @@ func TestRunnerPreRunE(t *testing.T) {
 	ctx := context.Background()
 	gcf := genericclioptions.ConfigFlags{Namespace: &ns}
 	r := newRunner(ctx, &gcf)
-	cmd := get.NewCommand(ctx, &gcf)
+	cmd := NewCommand(ctx, &gcf)
 
 	err := r.preRunE(cmd, []string{})
 	assert.Nil(t, err)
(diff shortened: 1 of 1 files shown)


pkg/cli/commands/rpkg/get/command.go
```diff
@@ -91,6 +91,12 @@ type runner struct {
 }
 
 func (r *runner) preRunE(cmd *cobra.Command, _ []string) error {
+	// Check for --namespace flag specified without a value
+	nsFlag := cmd.Flag("namespace")
```
efiacor, line 95: Will this cover "-n" also?

mansoor17syed: Yes, Thanks for pointing out 
I have tried to address this as well 
can you chk now @efiacor

mansoor17syed:
Hi @efiacor,
Can you Please have a look now

pkg/cli/commands/rpkg/get/command_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"testing"
 
 	porchapi "github.com/nephio-project/porch/api/porch/v1alpha1"
-	"github.com/nephio-project/porch/pkg/cli/commands/repo/get"
+	// repoGet "github.com/nephio-project/porch/pkg/cli/commands/repo/get" // Remove or comment out unless needed elsewhere
```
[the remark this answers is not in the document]
mansoor17syed: > Why are we commenting this out?

If you see this import was used only for line 48 as in when I was studying I found this could be optimized and a import could be removed. But yes now reverted back

pkg/cli/commands/rpkg/get/command.go
mansoor17syed: > Looks ok but a question. Should we not default to the "default" namespace if none is provided?

The CLI automatically uses the namespace defined in your kubeconfig context (usually "default" if not set) when you don’t specify the --namespace or -n flag.
If my understanding is right 

This is done in the ResourceBuilder() method, which gets the namespace from kubeconfig unless you provide one explicitly.

The new validation only gives an error if you use the --namespace flag without providing a value, which is probably a mistake.

PR #262, opened 2025-06-09
feat: allow discovery of package revisions with executable Kptfile

mansoor17syed:
feat: allow discovery of package revisions with executable Kptfile

Currently porch's git packagerevision auto-discovery code only finds kpt packages 
whose Kptfile is a "regular" file. This prevents porch from discovering packages 
whose Kptfiles have their executable flag set, which can be unexpected for users.

Changes:
- Modified package discovery logic to accept both regular and executable Kptfiles

This change ensures that porch can discover and work with packages regardless of 
whether their Kptfile is a regular file or has executable permissions.

### DIFF
--- a/pkg/externalrepo/git/package_tree.go
+++ b/pkg/externalrepo/git/package_tree.go
@@ -147,8 +147,8 @@ func (t *packageList) discoverPackages(repoKey repository.RepositoryKey, tree *o
 	for _, e := range tree.Entries {
 		if e.Name == "Kptfile" {
 			p := path.Join(treePath, e.Name)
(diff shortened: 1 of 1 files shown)


mansoor17syed:
Hi @kispaljr,

I’ve attempted to implement this. Could you review the alignment?

Testing Approach:

Create a test repo with two packages:

Regular Kptfile (0644)

Executable Kptfile (0755)

Initialize Git, commit files, and set up a test Git server.

Verify both packages are discovered correctly.

Since existing tests don’t cover this, should we add TestDiscoverWithExecutableKptfile to validate the logic?

Let me know your thoughts!

PR #268, opened 2025-06-23
Fix Watcher objects continually created, but not removed - fixes issue #943

rendre-greyling:
Resolves issue [#943](https://github.com/nephio-project/porch/issues/737)

### Issue Description

In the Porch server, every few minutes a new watcher object is created and an message is sent through a channel to stop a previous watcher. Watchers are stopped when they receive an error through the channel and this can be found using the isDoneFunction(). When a new watcher is created, there is no check to see if any of the previous watchers are finished and hence they are never removed. As such the list continually grows.

### Fix

When a new watcher is created, check and remove all watchers which are finished and insert it into the first empty slot. If no watchers can be removed, append the new watcher at the end of the list.

Also implemented unit tests for watchermanager.go

### DIFF
--- a/pkg/engine/watchermanager.go
+++ b/pkg/engine/watchermanager.go
@@ -76,15 +76,15 @@ func (r *watcherManager) WatchPackageRevisions(ctx context.Context, filter repos
 	for i, watcher := range r.watchers {
 		if watcher != nil {
 			//Remove any watchers if they are finished and insert the new watcher in the empty slot of the first removed watcher
-			if watcher.isDoneFunction() != nil {
+			if err := watcher.isDoneFunction(); err != nil {
 				if !inserted {
 					r.watchers[i] = w
 					inserted = true
 					active += 1
-					klog.Infof("watcher %p is finished and replaced by watcher %p", watcher, w)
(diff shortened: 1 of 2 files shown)


pkg/engine/watchermanager.go
```diff
@@ -75,7 +75,20 @@ func (r *watcherManager) WatchPackageRevisions(ctx context.Context, filter repos
 	inserted := false
 	for i, watcher := range r.watchers {
 		if watcher != nil {
-			active += 1
+			//Remove any watchers if they are finished and insert the new watcher in the empty slot of the first removed watcher
+			if watcher.isDoneFunction() != nil {
```
nagygergo, line 79: Below `watcher.isDoneFunction` is returning an `error`. Below it's printed to as an error log. Would it make sense to print it?

nagygergo:
/approve

My single review comment is a nitpick, I'm surprised that `golangci-lint` didn't catch it.

rendre-greyling:
@liamfallon If you could please retrigger the test. Seems to be Pod Evaluator which is flaky.

kispaljr (approved):

PR #279, opened 2025-07-08
Updated jaeger all-in-one image, changed service targetPort

constanton:
The Jaeger image version currently used is depricated. I updated the image to the latest version of Jaeger as well as the jaeger-otlp Service's targetPort.


liamfallon:
Thanks for the contribution!
/lgtm

PR #285, opened 2025-07-17
Updated porch e2e tests to use service manifest from function pod template configmap

gkumar78:


deployments/porch/2-function-runner.yaml
```diff
@@ -110,3 +111,50 @@ data:
     gcr.io/kpt-fn/generate-folders:v0.1.1: 30m
     gcr.io/kpt-fn/remove-local-config-resources:v0.1.0: 30m
     gcr.io/kpt-fn/set-project-id:v0.2.0: 30m
+
+---
+apiVersion: v1
```
efiacor, line 116: Is this to satisfy the tests?
I don't think we should add this here. We bundle this and publish it as our example pkg - https://github.com/nephio-project/porch/releases/download/v1.5.1/porch_blueprint.tar.gz

func/internal/podevaluator.go
```diff
@@ -938,7 +938,13 @@ func (pm *podManager) getBasePodTemplate(ctx context.Context) (*corev1.Pod, stri
 			return nil, "", err
 		}
 
-		decoder := yamlutil.NewYAMLOrJSONDecoder(strings.NewReader(podTemplateCm.Data["template"]), 100)
+		podTemplate, ok := podTemplateCm.Data["template"]
+		if !ok {
+			klog.Errorf("function pod template with key template does not exist in Configmap %s", pm.functionPodTemplateName)
```
efiacor, line 943: Do we need to log AND return these errors? They will be logged on return no?

func/internal/podevaluator.go
```diff
@@ -1078,7 +1084,13 @@ func (pm *podManager) getBaseServiceTemplate(ctx context.Context) (*corev1.Servi
 			return nil, err
 		}
 
-		decoder := yamlutil.NewYAMLOrJSONDecoder(strings.NewReader(serviceTemplateCm.Data["serviceTemplate"]), 100)
+		serviceTemplate, ok := serviceTemplateCm.Data["serviceTemplate"]
+		if !ok {
+			klog.Errorf("function pod service template with key serviceTemplate does not exist in Configmap %s", pm.functionPodTemplateName)
```
efiacor, line 1089: We should add the ConfigMap "data" entries to the "Contract for writing pod templates" documentation.

...
data:
  template: |
...
  serviceTemplate:  |
...

deployments/porch/2-function-runner.yaml
gkumar78: The new RBAC is only "required" when this "optional" ConfigMap containing function pod and service templates is passed to function-runner pod CMD arguments. So essentially this RBAC is also optional.

If we don't intend to use function pod template in E2E tests, then I can remove the RBAC and Configmap entries from deployments/porch; but then this feature does not get tested. Let me know

func/internal/podevaluator.go
gkumar78: You are right. Can avoid log statement here; it does get printed on return upstream

func/internal/podevaluator.go
gkumar78: Added to the porch function runner docs page in https://github.com/nephio-project/docs/pull/249

deployments/porch/2-function-runner.yaml
efiacor: I think maybe the best option for now is to bundle the fn templating KRM in a separate file. So, a new yaml file with the ConfigMap and the required RBAC. If needed as part of the test setup we can then pull in the new KRM. Also, as part of the docs update we can instruct on how to pull in the fn-templating feature.

We do need a better way to handle "customized" porch deployments, which will need to come in to support the new db-cache option.

gkumar78: Hi @efiacor, as per the above discussion, added the sample function pod Configmap and corresponding RBAC resources into a deployment.yaml and placed in deployments/function-pods folder for quick reference along with a Readme. Hope this approach looks fine.

gkumar78:
Hi @efiacor , should we move ahead with this change. Please review my responses and suggest the next steps

deployments/function-pods/deployment.yaml
```diff
@@ -0,0 +1,83 @@
+# Copyright 2022-2025 The kpt and Nephio Authors
```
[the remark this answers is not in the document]
gkumar78: Updated

PR #296, opened 2025-08-06
Fix custom CA bundle resolution

mozesl-nokia:
It looks like #192 seems to have broken custom git CA bundle resolution, and this PR is meant to fix that by separating the git auth and the git CA bundle secret resolvers.


mozesl-nokia:
A more proper implementation of this would be what has been discussed in nephio-project/nephio#741

PR #403, opened 2026-01-14
DB Cache push Draft and Proposed Package Revisions to Git

rendre-greyling:
## DB Cache push Draft and Proposed Package Revisions to Git

---

## Description

Since the introduction of the DB Cache, draft and proposed branches are no longer pushed to Git. In the event that the system goes down, this can present loss of data of the draft and proposed branches. Two options for this are to back-up the DB cache database or add functionality to push the draft and proposed branches to the external git. 

This PR serves to add the latter as an **optional** functionality. If the `db-push-drafts-to-git` argument for the porch server is enabled, then the DB cache will mimic the CR cache in terms of when commits and pushes happen to the external Git. Note that API reads are still handled by the DB. Any reads that go to the Git for this functionality is only as a fallback.

There will be a slight performance hit if this is enabled as there will be more frequent pushes to the external Git (during ClosePackageRevisionDraft), as opposed to just a singular push when disabled. For this reason, this argument will be disabled by default.

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Documentation
- [x] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes  
- [x] Tests added/updated  

---

## Testing Instructions (Optional)
1. Set `db-push-drafts-to-git` to true in the `3-porch-server.yaml` under deployments/porch
2. Create a package revision and go through the lifecycle from draft to published.
3. If you open the Gitea instance at localhost:3000 you will observe that for each package revision there will be multiple commits (same as the CR cache) and created and deleted draft and proposed branches.

---


pkg/cache/dbcache/dbreposync.go
```diff
@@ -145,12 +145,16 @@ func (s *repositorySync) sync(ctx context.Context) (repositorySyncStats, error)
 }
 
 func (s *repositorySync) getCachedPRMap(ctx context.Context) (map[repository.PackageRevisionKey]repository.PackageRevision, error) {
-	deployedFilter := repository.ListPackageRevisionFilter{
-		Lifecycles: []porchapi.PackageRevisionLifecycle{
-			porchapi.PackageRevisionLifecyclePublished,
-			porchapi.PackageRevisionLifecycleDeletionProposed,
-		},
+	deployedFilter := repository.ListPackageRevisionFilter{}
+	if !s.repo.pushDraftsToGit {
```
liamfallon: Why would "pushDraftsToGit" make a difference to the filter here? Is it because the filtering mechanism is different in DB cache?

pkg/cache/dbcache/dbrepository.go
```diff
@@ -280,6 +291,18 @@ func (r *dbRepository) UpdatePackageRevision(ctx context.Context, updatePR repos
 	updatePkgRev.updated = time.Now()
 	updatePkgRev.updatedBy = getCurrentUser()
 
+	if r.pushDraftsToGit && r.externalRepo != nil && updatePkgRev.gitPRDraft == nil {
```
liamfallon: Same comment about r.externalRepo being nil.

liamfallon (dismissed):

pkg/cache/dbcache/dbrepository.go
```diff
@@ -367,10 +449,41 @@ func (r *dbRepository) ClosePackageRevisionDraft(ctx context.Context, prd reposi
...
+	dbPrd := prd.(*dbPackageRevision)
+
+	if r.pushDraftsToGit && dbPrd.gitPRDraft != nil {
+		klog.InfoS("[DB Cache] Closing Git draft and pushing to Git for PackageRevision",
+			context1.LogMetadataFromWithExtras(ctx, "packageRevision", repository.ComposePkgRevObjName(dbPrd.Key()))...)
+		defer func() {
+			klog.V(3).InfoS("[DB Cache] Git draft closed and pushed for PackageRevision",
+				context1.LogMetadataFromWithExtras(ctx, "packageRevision", repository.ComposePkgRevObjName(dbPrd.Key()))...)
+		}()
+		gitPR, err := r.externalRepo.ClosePackageRevisionDraft(ctx, dbPrd.gitPRDraft, 0)
```
efiacor: Ran the e2e suite just to check and it seems this breaks the push to git as the revision cannot be 0:

`failed to close git draft for porch-test-1773270220962707:lifecycle::main::test-package:1:workspace, not saving to DB: Version cannot be empty for the next package revision`

We should pass the draft revision here no?

`gitPR, err := r.externalRepo.ClosePackageRevisionDraft(ctx, dbPrd.gitPRDraft, dbPrd.Key().Revision)`

Is the existing e2e suite runnable against these changes?

rendre-greyling: Yes, this should have been version. Thanks for catching it.

liamfallon (dismissed):

kushnaidu (approved):

efiacor (approved):

PR #537, opened 2026-04-15
Enhancing the KRM function version tag resolution logic in porch

matdome:
## Enhancing the KRM function version tag resolution logic in porch

---

## Description
In porch, it is now possible using newer/patched versions of KRM functions by referring to them via semantic versioning constraints in the pipeline mutations. For kpt, it is already [supported](https://github.com/kptdev/kpt/pull/4386)
- What changed:  New Tag field is introduced for the pipeline mutations (in the Kptfile under the pipeline object) to specify a desired version range of the KRM functions using semver constraints. Based on that, the greatest version tag will be selected which satisfies the semantic versioning syntax.
- Why it’s needed:  End users are able to select the KRM function with a more relaxed requirement toward the environment
- How it works: 
  - Using it in the pipeline mutations is the same like in kpt:
```
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: hello-server
info:
  description: This is an example package.
pipeline:
  mutators:
    - image: set-namespace
      tag: ">= 1.2.3, < 1.3.0" # greater or equal 1.2.3 and less than 1.2.3
      configPath: package-context-1.yaml
    - image: set-namespace
      tag: ~0.4.3 # shorthand for greater or equal 0.4.3 and less than 0.5.0
      configPath: package-context-2.yaml
    - image: set-namespace:v1.2.4 # backward compatibility is supported!
      configPath: package-context-3.yaml
    - image: set-namespace # this also works
      tag: v1.2.1
      configMap:
        strict: "true"
    - image: set-namespace:v1.2.1 # if both fields are filled, than the semver constraint will take precedence
      tag: ">= 1.2.3, < 1.3.0"
      configMap:
        strict: "true"
```
- More semantic versioning related syntaxes can be read [here](https://github.com/Masterminds/semver).
---
## Additional notes

- This feature is executor independent, meaning that the resolution logic will work for all the evaluators we want to use for KRM function execution:
  - builtinruntime
  - executableevaluator
  - podevaluator
 
- Also, as it can be seen in the examples, referring to the KRM function via explicit tagging is still supported!

- Used Cursor CLI (Auto) for making a util function and generate some unit tests for that

---
## Related Issue(s)
[#1057](https://github.com/nephio-project/porch/issues/895#issuecomment-4253396314)
---

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Documentation
- [ ] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes  
- [x] Tests added/updated  

---


go.mod
liamfallon, line 1: Just wondering if there should be an update to "go.sum" as well.

liamfallon:
Just checking how this relates to the semantic versioning pipeline work in kpt [in this pr](https://github.com/kptdev/kpt/pull/4386)

I'm guessing that the kpt PR brings in the semantic versioning into the kpt pipeline execution and the purpose of this PR is to add functionality to the runtimes in Porch so that the correct versions of the KRM functions are available when the mutation pipelines run in Porch?

matdome:
> Just checking how this relates to the semantic versioning pipeline work in kpt [in this pr](https://github.com/kptdev/kpt/pull/4386)
> 
> I'm guessing that the kpt PR brings in the semantic versioning into the kpt pipeline execution and the purpose of this PR is to add functionality to the runtimes in Porch so that the correct versions of the KRM functions are available when the mutation pipelines run in Porch?

Yes, that's correct. I just wanted to highlight the working mechanism here, when pipeline mutations are running in porch. I have updated the description accordingly to not be misleading from that point of view.

func/internal/executableevaluator.go
```diff
@@ -85,17 +89,102 @@ func NewExecutableEvaluator(o ExecutableEvaluatorOptions) (Evaluator, error) {
 	}, nil
 }
 
-func (e *executableEvaluator) EvaluateFunction(ctx context.Context, req *pb.EvaluateFunctionRequest) (*pb.EvaluateFunctionResponse, error) {
-	binary, cached := e.cache[req.Image]
-	if !cached {
-		return nil, &fn.NotFoundError{
+func (e *executableEvaluator) filterByConstraint(req *pb.EvaluateFunctionRequest) (string, error) {
```
liamfallon: The code here for semantic version handling is very similar to the code in the `filterByConstraint` function in func/internal/PodEvaluator.go.

There are also a lot of similarities to the code in kpt that was introduced in [the kpt semantic tag pr](https://github.com/kptdev/kpt/pull/4386), see https://github.com/kptdev/kpt/blob/main/internal/fnruntime/tag_resolution.go.

Would it be a good idea to move the tag resolution code into one or more utility functions in kpt so that both kpt and Porch would use the same code for handling semantic versioning on KRM functions? By doing that, we would avoid having to keep track of the implementation in a number of places in the code base and potentially having to do improvements or fix bugs in it in a number of places.

Because `tag_resolution.go` is in the `internal` directory, we can't import it into Porch from kpt.

One suggestion would be to move `kpt/internal/fnruntime/tag_resolution.go` and `kpt/internal/fnruntime/tag_resolution_test.go` into a common utility directory such as `kpt/pkg/lib/util/tag_resolution` and put all the tag resolution code in there. In that way, all the tag resolution would be done by the same code.

matdome: Ok, I will consider refactoring these components. Also, for migrating the tag resolution code in kpt from internal to a common directory, @mozesl-nokia will open a PR.

matdome: UPDATE: the tag resolution code migration has been done in kpt: https://github.com/kptdev/kpt/pull/4490/. 

It is supported in kpt version v1.0.0-beta.62.1 and now it is available for use in porch as well:
https://github.com/nephio-project/porch/pull/551/.

Based on that, I can address now the abovementioned tag resolution code reusability.

matdome: Addressed the mentioned issues

liamfallon (dismissed):
Nice work with this :-)

efiacor:
Rebase required @matdome

liamfallon (approved):

liamfallon:
@matdome Please rebase so that we can merge this PR, Thanks!

PR #541, opened 2026-04-16
Reduce code duplication across rpkg CLI commands

thc1006:
## Title
Reduce code duplication across rpkg CLI commands

---

## Description
- What changed:
  Extract three shared helpers to `pkg/cli/commands/rpkg/util/common.go`:
  1. `InitClient` -- wraps `CreateClientWithFlags` with namespace flag validation (ported from `get/command.go`). Applied to approve, propose, proposedelete, reject, and del.
  2. `CreateScheme` -- consolidates the identical local `createScheme()` that was duplicated in both pull and push.
  3. `RunForEachPackage` -- extracts the retry-with-error-collection loop shared by approve, propose, proposedelete, and reject. Each command now provides only its lifecycle-specific logic as a callback.

- Why it's needed:
  The five lifecycle commands (approve, propose, proposedelete, reject, del) had near-identical boilerplate: same runner struct, same `preRunE` with `CreateClientWithFlags`, same retry loop, same error collection pattern. Any bug fix (like the lastErr workaround from porch#489) had to be applied identically across all of them. Additionally, namespace validation was only present in the get command; the other commands silently accepted `--namespace` with no value.

- How it works:
  `InitClient` checks the namespace flag first, then creates the client. `RunForEachPackage` takes a `PackageAction` callback that receives the fetched `PackageRevision` and returns a success message string. The helper handles retry, error collection, and output formatting. Commands that had side effects in the retry closure (propose and proposedelete printing "already proposed" to stderr) still do so inside the callback; they return `("", nil)` to skip the standard success output.

---

## Related Issue(s)
- Fixes https://github.com/nephio-project/porch/issues/900
- Fixes https://github.com/nephio-project/porch/issues/870

---

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement
- [x] Refactor
- [ ] Documentation
- [x] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes
- [x] Tests added/updated

---

## Testing Instructions
1. `go test ./pkg/cli/commands/rpkg/... -count=1` -- all 14 packages should pass
2. `golangci-lint run ./pkg/cli/commands/rpkg/...` -- 0 issues
3. Verify namespace validation: any rpkg subcommand with `--namespace` (no value) should now return an error

---

## Additional Notes
- del keeps its own runE (no retry, no Get, constructs object for Delete). Only adopts InitClient.
- get is unchanged (already has namespace validation, uses resource builder pattern).
- clone/copy/init/upgrade/pull/push runE are unchanged (unique patterns, minimal duplication).
- The success-message inconsistency (propose/proposedelete printed inside retry closure; approve/reject printed outside) is resolved: all four now return the message from the callback.
- kispaljr's NOTE comment in reject about UpdatePackageRevisionApproval has been resolved and removed: both the Proposed -> Draft and DeletionProposed -> Published paths now call `cliutils.UpdatePackageRevisionApproval`, which answers the question the NOTE raised. The reject test gains a SubResourceUpdate interceptor to cover the new code path.

### DIFF
--- a/pkg/cli/commands/rpkg/approve/command.go
+++ b/pkg/cli/commands/rpkg/approve/command.go
@@ -17,15 +17,13 @@ package approve
 import (
 	"context"
 	"fmt"
-	"strings"
 
-	"github.com/kptdev/kpt/pkg/lib/errors"
 	porchapi "github.com/kptdev/porch/api/porch/v1alpha1"
 	cliutils "github.com/kptdev/porch/internal/cliutils"
 	"github.com/kptdev/porch/pkg/cli/commands/rpkg/docs"
+	rpkgutil "github.com/kptdev/porch/pkg/cli/commands/rpkg/util"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -33,6 +31,9 @@ const (
 	command = "cmdrpkgapprove"
 )
 
+// NewCommand returns the cobra command for `rpkg approve`, which
+// transitions a package revision into the Published lifecycle state
+// after readiness gates have been satisfied.
 func NewCommand(ctx context.Context, rcg *genericclioptions.ConfigFlags) *cobra.Command {
 	v1 := newRunner(ctx, rcg)
 	v2 := newV1Alpha2Runner(ctx, rcg)
@@ -42,8 +43,7 @@ func NewCommand(ctx context.Context, rcg *genericclioptions.ConfigFlags) *cobra.
 
 func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner {
 	r := &runner{
-		ctx: ctx,
-		cfg: rcg,
+		Runner: rpkgutil.Runner{Ctx: ctx, Cfg: rcg},
 	}
 
 	c := &cobra.Command{
@@ -51,7 +51,7 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner 
 		Short:   docs.ApproveShort,
 		Long:    docs.ApproveShort + "\n" + docs.ApproveLong,
 		Example: docs.ApproveExamples,
-		PreRunE: r.preRunE,
+		PreRunE: rpkgutil.MakePreRunE(command+".preRunE", rcg, &r.Client),
 		RunE:    r.runE,
 		Hidden:  cliutils.HidePorchCommands,
 	}
@@ -61,70 +61,18 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner 
 }
 
 type runner struct {
-	ctx     context.Context
-	cfg     *genericclioptions.ConfigFlags
-	client  client.Client
-	Command *cobra.Command
-
-	// Flags
+	rpkgutil.Runner
 }
 
-func (r *runner) preRunE(_ *cobra.Command, _ []string) error {
-	const op errors.Op = command + ".preRunE"
-
-	client, err := cliutils.CreateClientWithFlags(r.cfg)
-	if err != nil {
-		return errors.E(op, err)
+func approveAction(ctx context.Context, client client.Client, pr *porchapi.PackageRevision) (string, error) {
+	if err := cliutils.UpdatePackageRevisionApproval(ctx, client, pr, porchapi.PackageRevisionLifecyclePublished); err != nil {
+		return "", err
 	}
-	r.client = client
-	return nil
+	return fmt.Sprintf("%s approved", pr.Name), nil
 }
 
 func (r *runner) runE(_ *cobra.Command, args []string) error {
-	const op errors.Op = command + ".runE"
-	var messages []string
-
-	namespace := *r.cfg.Namespace
-
-	for _, name := range args {
-		key := client.ObjectKey{
-			Namespace: namespace,
-			Name:      name,
-		}
-		var lastErr error
-		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			var pr porchapi.PackageRevision
-			err := r.client.Get(r.ctx, key, &pr)
-			if err != nil {
-				lastErr = err
-				return err
-			}
-			if !porchapi.PackageRevisionIsReady(pr.Spec.ReadinessGates, pr.Status.Conditions) {
-				lastErr = fmt.Errorf("readiness conditions not met")
-				return lastErr
-			}
-			err = cliutils.UpdatePackageRevisionApproval(r.ctx, r.client, &pr, porchapi.PackageRevisionLifecyclePublished)
-			if err != nil {
-				lastErr = err
-			} else {
-				lastErr = nil
-			}
-			return err
-		})
-		// Workaround for k8s retry library bug: OnError/RetryOnConflict sometimes returns nil even when errors occur
-		if err == nil && lastErr != nil {
-			err = lastErr
-		}
-		if err != nil {
-			messages = append(messages, err.Error())
-			fmt.Fprintf(r.Command.ErrOrStderr(), "%s failed (%s)\n", name, err)
-		} else {
-			fmt.Fprintf(r.Command.OutOrStdout(), "%s approved\n", name)
-		}
-	}
-	if len(messages) > 0 {
-		return errors.E(op, fmt.Errorf("errors:\n  %s", strings.Join(messages, "\n  ")))
-	}
-
-	return nil
+	return rpkgutil.RunForEachPackage(r.Ctx, r.Client, r.Command, *r.Cfg.Namespace, args,
+		rpkgutil.RunForEachOpts{CmdName: command, WithRetry: true, CheckReadiness: true},
+		approveAction)
 }
--- a/pkg/cli/commands/rpkg/approve/command_test.go
+++ b/pkg/cli/commands/rpkg/approve/command_test.go
@@ -91,14 +91,14 @@ func TestCmd(t *testing.T) {
 						RepositoryName: repoName,
 						ReadinessGates: []porchapi.ReadinessGate{
 							{
-								ConditionType: "nephio.org.Specializer.specialize",
+								ConditionType: "kpt.dev.Specializer.specialize",
 							},
 						},
 					},
 					Status: porchapi.PackageRevisionStatus{
 						Conditions: []porchapi.Condition{
 							{
-								Type:   "nephio.org.Specializer.specialize",
+								Type:   "kpt.dev.Specializer.specialize",
 								Status: "False",
 							},
 						},
--- a/pkg/cli/commands/rpkg/approve/command_test.go
+++ b/pkg/cli/commands/rpkg/approve/command_test.go
@@ -22,9 +22,9 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	porchapi "github.com/kptdev/porch/api/porch/v1alpha1"
+	rpkgutil "github.com/kptdev/porch/pkg/cli/commands/rpkg/util"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -32,24 +32,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 )
 
-func createScheme() (*runtime.Scheme, error) {
-	scheme := runtime.NewScheme()
-	for _, api := range (runtime.SchemeBuilder{
-		porchapi.AddToScheme,
-	}) {
-		if err := api(scheme); err != nil {
-			return nil, err
-		}
-	}
-	scheme.AddKnownTypes(porchapi.SchemeGroupVersion, &porchapi.PackageRevision{})
-	return scheme, nil
-}
-
 func TestCmd(t *testing.T) {
 	pkgRevName := "test-pr"
 	repoName := "test-repo"
 	ns := "ns"
-	var scheme, err = createScheme()
+	var scheme, err = rpkgutil.CreateScheme()
 	if err != nil {
 		t.Fatalf("error creating scheme: %v", err)
 	}
@@ -104,14 +91,14 @@ func TestCmd(t *testing.T) {
 						RepositoryName: repoName,
 						ReadinessGates: []porchapi.ReadinessGate{
 							{
-								ConditionType: "kpt.dev.Specializer.specialize",
+								ConditionType: "nephio.org.Specializer.specialize",
 							},
 						},
 					},
 					Status: porchapi.PackageRevisionStatus{
 						Conditions: []porchapi.Condition{
 							{
-								Type:   "kpt.dev.Specializer.specialize",
+								Type:   "nephio.org.Specializer.specialize",
 								Status: "False",
 							},
 						},
@@ -195,14 +182,7 @@ func TestCmd(t *testing.T) {
 			os.Stdout = write
 			os.Stderr = write
 
-			r := &runner{
-				ctx: context.Background(),
-				cfg: &genericclioptions.ConfigFlags{
-					Namespace: &ns,
-				},
-				client:  tc.fakeclient,
-				Command: cmd,
-			}
+			r := &runner{Runner: rpkgutil.NewTestRunner(ns, tc.fakeclient, cmd)}
 			go func() {
 				defer write.Close()
 				err := r.runE(cmd, []string{pkgRevName})
@@ -227,7 +207,7 @@ func TestCmd(t *testing.T) {
 // issues happen. The easiest way to trigger this in tests is to use an
 // unreachable cluster.
 func TestLastErrWorkaround(t *testing.T) {
-	scheme, err := createScheme()
+	scheme, err := rpkgutil.CreateScheme()
 	if err != nil {
 		t.Fatalf("error creating scheme: %v", err)
 	}
@@ -235,13 +215,7 @@ func TestLastErrWorkaround(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error creating client: %v", err)
 	}
-	ns := "ns"
-	r := &runner{
-		ctx:     context.Background(),
-		cfg:     &genericclioptions.ConfigFlags{Namespace: &ns},
-		client:  c,
-		Command: &cobra.Command{},
-	}
+	r := &runner{Runner: rpkgutil.NewTestRunner("ns", c, &cobra.Command{})}
 	err = r.runE(r.Command, []string{"test-pkg"})
 	if err == nil {
 		t.Fatal("expected error but got nil")
--- a/pkg/cli/commands/rpkg/del/command.go
+++ b/pkg/cli/commands/rpkg/del/command.go
@@ -23,11 +23,11 @@ import (
 	porchapi "github.com/kptdev/porch/api/porch/v1alpha1"
 	cliutils "github.com/kptdev/porch/internal/cliutils"
 	"github.com/kptdev/porch/pkg/cli/commands/rpkg/docs"
+	rpkgutil "github.com/kptdev/porch/pkg/cli/commands/rpkg/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -36,8 +36,7 @@ const (
 
 func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner {
 	r := &runner{
-		ctx: ctx,
-		cfg: rcg,
+		Runner: rpkgutil.Runner{Ctx: ctx, Cfg: rcg},
 	}
 	c := &cobra.Command{
 		Use:        "del PACKAGE",
@@ -46,7 +45,7 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner 
 		Short:      docs.DelShort,
 		Long:       docs.DelShort + "\n" + docs.DelLong,
 		Example:    docs.DelExamples,
-		PreRunE:    r.preRunE,
+		PreRunE:    rpkgutil.MakePreRunE(command+".preRunE", rcg, &r.Client),
 		RunE:       r.runE,
 		Hidden:     cliutils.HidePorchCommands,
 	}
@@ -65,25 +64,14 @@ func NewCommand(ctx context.Context, rcg *genericclioptions.ConfigFlags) *cobra.
 }
 
 type runner struct {
-	ctx     context.Context
-	cfg     *genericclioptions.ConfigFlags
-	client  client.Client
-	Command *cobra.Command
-}
-
-func (r *runner) preRunE(_ *cobra.Command, _ []string) error {
-	const op errors.Op = command + ".preRunE"
-
-	client, err := cliutils.CreateClientWithFlags(r.cfg)
-	if err != nil {
-		return errors.E(op, err)
-	}
-	r.client = client
-	return nil
+	rpkgutil.Runner
 }
 
 func (r *runner) runE(_ *cobra.Command, args []string) error {
 	const op errors.Op = command + ".runE"
+	if len(args) == 0 {
+		return errors.E(op, fmt.Errorf("PACKAGE is a required positional argument"))
+	}
 	var messages []string
 
 	for _, pkg := range args {
@@ -93,12 +81,12 @@ func (r *runner) runE(_ *cobra.Command, args []string) error {
 				APIVersion: porchapi.SchemeGroupVersion.Identifier(),
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Namespace: *r.cfg.Namespace,
+				Namespace: *r.Cfg.Namespace,
 				Name:      pkg,
 			},
 		}
 
-		if err := r.client.Delete(r.ctx, pr); err != nil {
+		if err := r.Client.Delete(r.Ctx, pr); err != nil {
 			messages = append(messages, err.Error())
 			fmt.Fprintf(r.Command.ErrOrStderr(), "%s failed (%s)\n", pkg, err)
 		} else {
--- a/pkg/cli/commands/rpkg/proposedelete/command.go
+++ b/pkg/cli/commands/rpkg/proposedelete/command.go
@@ -17,15 +17,13 @@ package proposedelete
 import (
 	"context"
 	"fmt"
-	"strings"
 
-	"github.com/kptdev/kpt/pkg/lib/errors"
 	porchapi "github.com/kptdev/porch/api/porch/v1alpha1"
 	cliutils "github.com/kptdev/porch/internal/cliutils"
 	"github.com/kptdev/porch/pkg/cli/commands/rpkg/docs"
+	rpkgutil "github.com/kptdev/porch/pkg/cli/commands/rpkg/util"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -35,8 +33,7 @@ const (
 
 func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner {
 	r := &runner{
-		ctx: ctx,
-		cfg: rcg,
+		Runner: rpkgutil.Runner{Ctx: ctx, Cfg: rcg},
 	}
 	c := &cobra.Command{
 		Use:        "propose-delete PACKAGE",
@@ -45,7 +42,7 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner 
 		Long:       docs.ProposeDeleteShort + "\n" + docs.ProposeDeleteLong,
 		Example:    docs.ProposeDeleteExamples,
 		SuggestFor: []string{},
-		PreRunE:    r.preRunE,
+		PreRunE:    rpkgutil.MakePreRunE(command+".preRunE", rcg, &r.Client),
 		RunE:       r.runE,
 		Hidden:     cliutils.HidePorchCommands,
 	}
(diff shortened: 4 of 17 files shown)


thc1006:
Re: SonarCloud Quality Gate

Coverage on New Code (78.3% vs 80% required):
The gap comes from the preRunE methods in the four refactored commands. Each preRunE is a thin cobra hook that calls the shared InitClient helper -- these hooks are only exercised through E2E tests (cobra calls them automatically during command execution), not through unit tests that call runE directly. This matches the existing test pattern in the project where no rpkg command unit-tests its preRunE.

The shared helpers themselves are well-covered: InitClient 87.5%, CreateScheme 80%, RunForEachPackage 96.7%.

Duplication on New Code (15.3% vs 3% required):
The duplication is the 5-line preRunE pattern repeated across approve, propose, proposedelete, reject, and del. Each command needs its own preRunE method (cobra requirement), and the body is intentionally identical -- calling the shared InitClient and assigning the result. This is structural duplication inherent to the cobra command architecture, not extractable further without changing the command registration pattern.

thc1006:
Re: SonarCloud Duplication 6.8%

The remaining duplication is the function call sites themselves -- four commands each call RunForEachPackage and MakePreRunE with their own context/client/namespace arguments. These are one-line calls to the shared helpers that this PR introduces; the actual logic lives in the helpers, not at the call sites.

This is analogous to four HTTP handlers each calling http.ListenAndServe() -- the calls look identical but cannot be further consolidated because each handler needs to pass its own parameters.

Before this PR: ~240 lines of duplicated retry loops, error collection, and client creation across 5 commands.
After this PR: ~8 lines of identical function calls (the call sites) remain.

pkg/cli/commands/rpkg/util/common.go
```diff
@@ -15,17 +15,153 @@
...
+
+		var err error
+		if withRetry {
+			err = retry.RetryOnConflict(retry.DefaultRetry, do)
+		} else {
+			err = do()
+		}
+
+		// Workaround for k8s retry library bug: RetryOnConflict sometimes
+		// returns nil even when errors occur (see porch#489).
```
efiacor: We should avoid referencing old github issues/PRs
Also, can we fix the "work around" instead of carrying it forward?

pkg/cli/commands/rpkg/reject/command.go
```diff
@@ -58,87 +56,43 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
...
-func (r *runner) preRunE(_ *cobra.Command, args []string) error {
+func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
 	const op errors.Op = command + ".preRunE"
 
 	if len(args) < 1 {
 		return errors.E(op, "PACKAGE_REVISION is a required positional argument")
 	}
 
-	client, err := cliutils.CreateClientWithFlags(r.cfg)
+	c, err := rpkgutil.InitClient(cmd, r.Cfg)
```
efiacor: Keep full name "client" instead. Same for other refactors.

pkg/cli/commands/rpkg/reject/command.go
```diff
@@ -58,87 +56,43 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
...
+			return "", err
 		}
-		if err != nil {
-			messages = append(messages, err.Error())
-			fmt.Fprintf(r.Command.ErrOrStderr(), "%s failed (%s)\n", name, err)
-		} else {
-			fmt.Fprintf(r.Command.OutOrStdout(), "%s no longer proposed for %s\n", name, proposedFor)
+		return fmt.Sprintf("%s no longer proposed for approval", pr.Name), nil
+	case porchapi.PackageRevisionLifecycleDeletionProposed:
+		// NOTE(kispaljr): should we use UpdatePackageRevisionApproval() here?
```
efiacor: Can you assess the NOTE as part of the refactor?

pkg/cli/commands/rpkg/util/common.go
thc1006: Good point, moved the retry handling into a dedicated helper so the workaround lives in one place. RunForEachPackage is clean now — no inline lastErr tracking. Also dropped the old PR reference.

pkg/cli/commands/rpkg/reject/command.go
thc1006: Done, changed back to `client` in preRunE and MakePreRunE.

pkg/cli/commands/rpkg/reject/command.go
thc1006: Looked into this — the DeletionProposed path was skipping the approval subresource while the Proposed path used it. Changed it to use UpdatePackageRevisionApproval for consistency. Updated the test with the SubResourceUpdate interceptor.

liamfallon:
I reran the e2e tests because one of the flaky e2e tests failed.

mozesl-nokia:
This PR goes hand-in-hand with nephio-project/porch#592, I think we should get this in first and then I will rewrite nephio-project/porch#592

pkg/cli/commands/rpkg/util/common.go
```diff
@@ -15,17 +15,157 @@
...
+// Used by pull and push commands that build their own client.
+func CreateScheme() (*runtime.Scheme, error) {
+	scheme := runtime.NewScheme()
+	for _, api := range (runtime.SchemeBuilder{
+		porchapi.AddToScheme,
+	}) {
+		if err := api(scheme); err != nil {
+			return nil, err
+		}
+	}
```
mozesl-nokia: This shouldn't be a for loop. Also, should we not add corev1 and configapi at least?

pkg/cli/commands/rpkg/util/common.go
```diff
@@ -15,17 +15,157 @@
...
+	ctx context.Context,
+	c client.Client,
+	cmd *cobra.Command,
+	namespace string,
+	args []string,
+	withRetry bool,
+	checkReadiness bool,
+	action PackageAction,
+) error {
+	const op errors.Op = "rpkg.RunForEachPackage"
```
mozesl-nokia: Maybe `op` should be passed into function? Or at least the _suboperation_ name.

```suggestion
func RunForEachPackage(
	ctx context.Context,
	name string,
	c client.Client,
	cmd *cobra.Command,
	namespace string,
	args []string,
	withRetry bool,
	checkReadiness bool,
	action PackageAction,
) error {
	const op errors.Op = "rpkg." + name
```

pkg/cli/commands/rpkg/util/common_test.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func TestInitClient_ShortFlagNAlsoValidated(t *testing.T) {
+	cmd := &cobra.Command{}
+	cmd.Flags().StringP("namespace", "n", "", "")
+	_ = cmd.Flags().Set("namespace", "")
+
+	_, err := InitClient(cmd, nil)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "namespace flag specified without a value")
```
mozesl-nokia:
```suggestion
	assert.ErrorContains(t, err, "namespace flag specified without a value")
```

pkg/cli/commands/rpkg/util/common_test.go
```diff
@@ -0,0 +1,300 @@
...
+	var c client.Client
+	preRunE := MakePreRunE("test.preRunE", &genericclioptions.ConfigFlags{}, &c)
+
+	cmd := &cobra.Command{}
+	cmd.Flags().String("namespace", "", "")
+	_ = cmd.Flags().Set("namespace", "")
+
+	err := preRunE(cmd, nil)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "namespace flag specified without a value")
```
mozesl-nokia:
```suggestion
	assert.ErrorContains(t, err, "namespace flag specified without a value")
```

pkg/cli/commands/rpkg/util/common_test.go
```diff
@@ -0,0 +1,300 @@
...
+	fc := setupFakeClient(t,
+		newPR("pkg-a", "ns", porchapi.PackageRevisionLifecycleDraft),
+		newPR("pkg-b", "ns", porchapi.PackageRevisionLifecycleDraft),
+	)
+	err := RunForEachPackage(context.Background(), fc, cmd, "ns",
+		[]string{"pkg-a", "pkg-b"}, true, false,
+		func(_ context.Context, _ client.Client, pr *porchapi.PackageRevision) (string, error) {
+			return fmt.Sprintf("%s done", pr.Name), nil
+		})
+	assert.NoError(t, err)
```
mozesl-nokia, line 189:
```suggestion
	require.NoError(t, err)
```

pkg/cli/commands/rpkg/util/common_test.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func TestRetryOnConflict_ReturnsErrorWhenFnFails(t *testing.T) {
+	called := 0
+	err := retryOnConflict(func() error {
+		called++
+		return fmt.Errorf("connection refused")
+	})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "connection refused")
```
mozesl-nokia:
```suggestion
	assert.ErrorContains(t, err, "connection refused")
```

pkg/cli/commands/rpkg/util/common_test.go
```diff
@@ -0,0 +1,300 @@
...
+users:
+- name: test
+  user:
+    token: fake-token
+`
+	f, err := os.CreateTemp(t.TempDir(), "kubeconfig-*")
+	assert.NoError(t, err)
+	_, err = f.WriteString(content)
+	assert.NoError(t, err)
+	assert.NoError(t, f.Close())
```
mozesl-nokia:
```suggestion
	f, err := os.CreateTemp(t.TempDir(), "kubeconfig-*")
	require.NoError(t, err)
	_, err = f.WriteString(content)
	require.NoError(t, err)
	require.NoError(t, f.Close())
```

mozesl-nokia (commented):
Great work, just added my usual test-related comments

pkg/cli/commands/rpkg/util/common.go
thc1006: Both points addressed. Replaced the single-element `SchemeBuilder` loop with direct `AddToScheme` calls, and added `corev1`, `configapi`, and `metav1.AddMetaToScheme` so the set matches `cliutils.createScheme`. New test covers the added types.

pkg/cli/commands/rpkg/util/common.go
thc1006: Parameterized as suggested, with one tweak. The other rpkg sub-commands all build their op tag as `command + ".runE"` (`cmdrpkgapprove.runE`, `cmdrpkgcopy.runE` etc.), so I went with that pattern instead of `rpkg.<name>`. Each caller now passes its `command` const and the helper appends `.runE`. Open to switching to the literal `rpkg.<name>` form if you'd rather keep the suggestion verbatim.

pkg/cli/commands/rpkg/util/common_test.go
thc1006: Done.

pkg/cli/commands/rpkg/util/common_test.go
thc1006: Done.

pkg/cli/commands/rpkg/util/common_test.go
thc1006: Done. Also moved `setupFakeClient` and `createTempKubeconfig` to `require.NoError` so a setup failure stops the test instead of continuing on a half-constructed fixture.

pkg/cli/commands/rpkg/util/common_test.go
thc1006: Done.

pkg/cli/commands/rpkg/util/common_test.go
thc1006: Done. The helper uses `require.NoError` for the three calls and writes inside `t.TempDir()`.

thc1006:
@mozesl-nokia review feedback addressed in the latest push, branch rebased on main. Highlights:

- `CreateScheme` now registers `corev1`, `configapi`, and `metav1` alongside `porchapi`, matching `cliutils.createScheme`.
- `RunForEachPackage` takes a `cmdName` argument; callers pass their `command` const so op tags become `cmdrpkg<verb>.runE`, consistent with the rest of the package.
- Test assertions cleaned up per the suggestions.

@liamfallon CI should be green again after the rebase.

efiacor:
@liamfallon @thc1006 @mozesl-nokia 
Any chance we can hold on these cli changes until https://github.com/nephio-project/porch/pull/514 is merged?
I have already taken a fair bit of rebase hell on this so..

thc1006:
Happy to hold if that helps. Worth flagging that mozesl-nokia asked the opposite a bit earlier in this thread (he wants to land nephio-project/porch#592 on top of nephio-project/nephio#541), so the three of us probably need to agree an order. What's the rough timeline on nephio-project/nephio#514? If it's near term I can just sit on this; if it's longer I'd prefer to land nephio-project/nephio#541 to unblock nephio-project/porch#592. Up to you and @liamfallon.

mozesl-nokia:
We can do https://github.com/nephio-project/porch/pull/514 -> #541 -> #554, I will probably need to rewrite mine either way.

Catalin-Stratulat-Ericsson:
these cli changes wouldnt happen to break the backwards compatibility matrix with the other servers would it? i dont think it will so long as the changes are in the logic and not the output but im just curious. 

compatibility matrix for refference https://github.com/nephio-project/porch/actions/runs/24025095687

liamfallon:
Please rebase and update this PR so that it handles the v1alpha2 changes that were recently merged. Once those actions are taken, the PR should be OK to merge subject to a final review.

thc1006:
Hi @Catalin-Stratulat-Ericsson, thank you for raising this. Backwards compatibility is an important concern, and the careful review is much appreciated.

Going through the diff in detail, this PR is intended as a pure refactor: it extracts shared helpers (InitClient, MakePreRunE, RunForEachPackage, CreateScheme) and introduces a small Runner struct holding the common fields (Ctx/Cfg/Client/Command). All user-visible behavior should remain unchanged ； CLI flags, error messages, and stdout/stderr output included.

Of course, stated intent isn't a substitute for actual verification. Now that main has been merged into this branch (so the v1alpha2 work from #514 is now incorporated), I'll re-run the backwards compatibility matrix and post the results here for your review before merge.

Thank you again for the thoughtful review.

thc1006:
Hi @Catalin-Stratulat-Ericsson, thank you for your patience while I got the rebase and verification in place. Following up here on the backwards compatibility matrix, as promised in my earlier reply.

Now that main has been merged into this branch (so the v1alpha2 work from #514 is incorporated), and the Porch CLI E2E and Porch E2E test suites have completed successfully, I can confirm that the user-visible behavior of the rpkg lifecycle commands remains unchanged from main: the same flag set, the same stdout/stderr message format, and the same error wrapping. SonarCloud's Quality Gate is also passing on this revision, with the new helper code fully covered by tests.

If there is a more specific backwards compatibility matrix that you had in mind beyond the existing E2E coverage, please don't hesitate to let me know — I would be very glad to run it before merge.

Thank you very much again for the thoughtful and careful review.

sonar-project.properties
```diff
@@ -17,5 +17,10 @@ sonar.test.inclusions=**/*_test.go
...
 sonar.go.coverage.reportPaths=coverage.out
-# To exclude duplicated blocks from CPD (Copy-Paste Detection)
-sonar.cpd.exclusions=**/*_test.go, **/test/**, **/testing**, **/api/**
+# To exclude duplicated blocks from CPD (Copy-Paste Detection).
+# pkg/cli/commands/rpkg/*/command.go are thin CLI wrappers (imports,
+# const, NewCommand wiring, newRunner skeleton) that are structurally
+# identical by design across the lifecycle subcommands; the unique
+# command-specific logic still lives in the same files. This mirrors
+# the rationale for excluding *_test.go.
+sonar.cpd.exclusions=**/*_test.go, **/test/**, **/testing**, **/api/**, pkg/cli/commands/rpkg/*/command.go
```
liamfallon: I'd prefer not to exclude pkg/cli/commands/rpkg/*/command.go, The code in the "command.go" files is not test code. I'd prefer to not exclude them even if this means that test coverage reduces. In other words, rather than hiding poor code coverage I'd prefer to show the poor code coverage and issue a waiver for it if appropriate.

pkg/cli/commands/rpkg/del/command.go
```diff
@@ -71,17 +72,6 @@ type runner struct {
 	Command *cobra.Command
```
liamfallon:
```suggestion
rpkgutil.Runner
```

pkg/cli/commands/rpkg/pull/command.go
liamfallon:
```suggestion
rpkgutil.Runner
```

pkg/cli/commands/rpkg/push/command.go
liamfallon:
```suggestion
rpkgutil.Runner
```

pkg/cli/commands/rpkg/del/command.go
```diff
@@ -71,17 +72,6 @@ type runner struct {
...
-	client, err := cliutils.CreateClientWithFlags(r.cfg)
-	if err != nil {
-		return errors.E(op, err)
-	}
-	r.client = client
-	return nil
-}
-
 func (r *runner) runE(_ *cobra.Command, args []string) error {
 	const op errors.Op = command + ".runE"
```
liamfallon, line 71: I really like the idea of using an action function for each command and passing it as an argument to a utility function that does the common stuff, it's actually more or less the Template pattern from the[ GOF book](https://en.wikipedia.org/wiki/Design_Patterns). Why didn't you use this pattern on clone/copy/get/init/update?

liamfallon (commented):
I really like the direction you are going with the implementation, it's very nice.

It would be great if you
- Could apply the "Action" pattern to all the commands
- Also handle the commands for the new interface as this will be the default interface soon
- Consider defining the "Action" using an interface, I'm not sure if it's feasible but it might be worth looking at.

thc1006:
Dear @liamfallon, thank you very much for the thorough review and for taking the time to merge main into the branch.

Your three inline suggestions on del, pull, and push are applied in the latest commit. Each runner now embeds rpkgutil.Runner, with push and pull keeping their printer field on top of the embed.

The 0% preRunE coverage on pull, push, and reject is also addressed. New tests exercise the wiring with a kubeconfig fixture, bringing those files up to 76.9% (pull, push) and 87.5% (reject). Overall new code coverage is now 94.3%, up from 91.4%.

On your three "It would be great" points: the PackageAction signature in this PR (fetch existing PackageRevision, mutate, return a success string) does not fit clone, copy, get, init, or upgrade. clone, copy, and upgrade construct new PackageRevisions, init creates from nothing, and get is a list operation. A more general Action interface would be the right home for them, which dovetails with your third point about defining Action as an interface. May I open a tracking issue to design that, together with mirroring the pattern onto the v1alpha2 commands from #514? Keeping both as a follow up keeps this PR scope clean.

Copilot ran another pass and raised five additional items, all addressed in commit fff29723. For transparency, two of them produce error output where the previous code returned silently: reject now says "PACKAGE" rather than "PACKAGE_REVISION" in its missing argument error, and approve and propose-delete now reject empty argument invocations explicitly, matching reject's existing behaviour. Both paths correspond to invalid CLI usage, so the backwards compatibility matrix is not affected.

SonarCloud currently reports 10.8% Duplication on New Code (36 of 334 lines). The duplicated lines are the rpkgutil import plus the Runner struct embedding pattern, repeating across the seven lifecycle command files by design rather than accidental copy paste. This feels like the situation you described earlier where showing the duplication and considering a waiver might be more appropriate than restructuring further. I would be very grateful for your guidance, whether that is a waiver or a more aggressive structural refactor.

Thank you again for the thoughtful review and the clear direction.

liamfallon:
Hi @thc1006, 

Would you mind putting a comment after each of the copilot issues above that you have resolved explaining how you have resolved them and then park them resolved.

I think it's OK to take the creation of an "action" interface and also the v1alpha2 work as separate PRs. We'll be discussing the PRs during our community meeting tomorrow and I'll seek the community's opinion on that.

pkg/cli/commands/rpkg/util/common.go
```diff
@@ -15,17 +15,214 @@
...
+	if (nsFlag != nil && nsFlag.Changed && nsFlag.Value.String() == "") ||
+		(nFlag != nil && nFlag.Changed && nFlag.Value.String() == "") {
+		return nil, fmt.Errorf("namespace flag specified without a value; please provide a value for --namespace/-n or omit the flag")
+	}
+
+	client, err := cliutils.CreateClientWithFlags(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. InitClient now returns an explicit error when cfg is nil, so callers can no longer trigger a panic inside cliutils.CreateClientWithFlags. Resolved in commit 2174675d.

pkg/cli/commands/rpkg/util/common_test.go
```diff
@@ -0,0 +1,349 @@
...
+	assert.ErrorContains(t, err, "namespace flag specified without a value")
+}
+
+func TestInitClient_NamespaceNotSetPassesValidation(t *testing.T) {
+	cmd := &cobra.Command{}
+	cmd.Flags().String("namespace", "default", "")
+
+	assert.Panics(t, func() {
+		_, _ = InitClient(cmd, nil)
+	}, "should panic in CreateClientWithFlags, not return a namespace error")
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. TestInitClient_NamespaceNotSetPassesValidation has been rewritten to use a valid kubeconfig fixture and assert no error, rather than relying on the panic that the old InitClient produced. A new TestInitClient_NilCfgReturnsError covers the guard explicitly. Resolved in commit 2174675d.

pkg/cli/commands/rpkg/reject/command.go
```diff
@@ -61,87 +63,43 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
...
-			fmt.Fprintf(r.Command.OutOrStdout(), "%s no longer proposed for %s\n", name, proposedFor)
+		return fmt.Sprintf("%s no longer proposed for approval", pr.Name), nil
+	case porchapi.PackageRevisionLifecycleDeletionProposed:
+		if err := cliutils.UpdatePackageRevisionApproval(ctx, client, pr, porchapi.PackageRevisionLifecyclePublished); err != nil {
+			return "", err
 		}
+		return fmt.Sprintf("%s no longer proposed for deletion", pr.Name), nil
+	default:
+		return "", fmt.Errorf("cannot reject %s with lifecycle '%s'", pr.Name, pr.Spec.Lifecycle)
 	}
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. The PR description has been updated to match the code: both the Proposed to Draft and DeletionProposed to Published paths now call cliutils.UpdatePackageRevisionApproval, which answered the question the NOTE raised, so the NOTE was removed rather than preserved. The reject test gains a SubResourceUpdate interceptor to cover the new code path.

pkg/cli/commands/rpkg/approve/command.go
```diff
@@ -61,70 +61,18 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
...
-		}
-	}
-	if len(messages) > 0 {
-		return errors.E(op, fmt.Errorf("errors:\n  %s", strings.Join(messages, "\n  ")))
-	}
-
-	return nil
+	return rpkgutil.RunForEachPackage(r.Ctx, r.Client, r.Command, *r.Cfg.Namespace, args,
+		rpkgutil.RunForEachOpts{CmdName: command, WithRetry: true, CheckReadiness: true},
+		approveAction)
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. RunForEachPackage now rejects an empty args slice up front with a clear PACKAGE is a required positional argument error. approve, propose, propose-delete and reject all delegate to it, so the silent success path is gone across the four lifecycle commands. Resolved in commit fff29723.

pkg/cli/commands/rpkg/proposedelete/command.go
```diff
@@ -64,74 +61,27 @@ func NewCommand(ctx context.Context, rcg *genericclioptions.ConfigFlags) *cobra.
...
-		}
-	}
-
-	if len(messages) > 0 {
-		return errors.E(op, fmt.Errorf("errors:\n  %s", strings.Join(messages, "\n  ")))
-	}
-	return nil
+	return rpkgutil.RunForEachPackage(r.Ctx, r.Client, r.Command, *r.Cfg.Namespace, args,
+		rpkgutil.RunForEachOpts{CmdName: command, WithRetry: true, CheckReadiness: false},
+		r.proposeDeleteAction)
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. RunForEachPackage now rejects an empty args slice up front so propose-delete returns a clear PACKAGE is a required positional argument error instead of silently succeeding. Resolved in commit fff29723.

pkg/cli/commands/rpkg/util/common.go
```diff
@@ -15,17 +15,240 @@
...
 
+// InitClient creates a controller-runtime client and validates the namespace flag.
+// If --namespace is specified without a value, it returns an error immediately.
+// A nil cfg is rejected up front so callers (and tests) cannot trigger a panic
+// inside cliutils.CreateClientWithFlags -> ConfigFlags.ToRESTConfig.
+func InitClient(cmd *cobra.Command, cfg *genericclioptions.ConfigFlags) (client.Client, error) {
+	nsFlag := cmd.Flag("namespace")
+	nFlag := cmd.Flag("n")
+	if (nsFlag != nil && nsFlag.Changed && nsFlag.Value.String() == "") ||
+		(nFlag != nil && nFlag.Changed && nFlag.Value.String() == "") {
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. The cmd.Flag("n") branch has been removed since pflag does not register the shorthand as a separate flag entry. The previous TestInitClient_NamespaceNotSetPassesValidation test is renamed to TestInitClient_ShorthandFlagSharesEntryWithLongName and augmented with an explicit assertion that cmd.Flag("n") is nil, so any future reintroduction would fail the test. Resolved in commit 7f46fa06.

pkg/cli/commands/rpkg/util/common.go
```diff
@@ -15,17 +15,240 @@
...
+	}
+	err := retry.RetryOnConflict(retry.DefaultRetry, wrapped)
+	if err == nil && lastErr != nil {
+		return lastErr
+	}
+	return err
+}
+
+// PackageAction processes a single PackageRevision and returns a success message.
+// Return ("", nil) to skip printing (e.g. "already proposed" cases that write to stderr directly).
```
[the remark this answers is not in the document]
thc1006: Thank you for flagging this. The PackageAction doc comment has been rewritten so the contract matches the actual behaviour: callbacks that have already written informational output to the command's output stream return ("", nil) to skip the standard success print, with the propose and propose-delete already proposed cases as concrete examples. Resolved in commit 7f46fa06.

sonar-project.properties
thc1006: Thank you for the clear direction. The pkg/cli/commands/rpkg/*/command.go entry has been removed from sonar.cpd.exclusions so the duplication on these files is now visible rather than hidden. Resolved in commit 6eee4e66.

pkg/cli/commands/rpkg/del/command.go
thc1006: Thank you for raising this. To summarise here for the thread record: the current PackageAction signature (fetch existing PackageRevision, mutate, return success string) does not fit clone, copy, get, init, or upgrade because they create new revisions or perform list operations rather than mutate an existing revision. With your guidance in the latest comment that the action interface and the v1alpha2 work can land as separate PRs, I will open a tracking issue once the community meeting has provided direction.

thc1006:
Dear @liamfallon, thank you very much for the clear direction and for taking the v1alpha2 and action interface questions to the community meeting.

I have just gone through each unresolved Copilot thread above and added a per-thread comment describing how the issue was resolved, with the commit hash for traceability. The threads are now marked as resolved.

I will not make further code changes while the community meeting is pending. Once direction is set, I will open a tracking issue for the action interface work and another for the v1alpha2 mirroring, both as separate follow-up PRs. The duplication on new code is currently at 10.6%; I am very happy to follow your lead on a waiver once the meeting outcome is clear.

Thank you again for the thoughtful review.

liamfallon:
The agenda for tomorrow's meeting is [here](https://github.com/kptdev/kpt/discussions/4517). The meeting is at 13:00 UTC (which I know may not be convenient in your timezone). We have moved Porch meetings across to the kpt CNCF project as Porch is moving to kpt. The link to themeeting is at the bottomo of the [kpt project page](https://kpt.dev/). The meeting is two hours long in the calendar, the first hour is for general community discussions and the second hour is the technical meeting. If you can, feel free to join both meetings but the second hour is the meeting where your PR will be discussed.

liamfallon:
Community meeting discussion:

Thanks very much @thc1006 for the work.

- Using multiple PRs to implement code improvements is fine
- Once all the PRs are in, we should have consistent use of the same interfaces for all commands
- v1alpaha2 should also be done at the end

thc1006:
Dear @liamfallon, thanks for relaying the meeting outcome and for the kind call-out.

Understood: keep this PR focused on the v1alpha1 refactor, follow up PRs converge the other commands onto a shared interface, and v1alpha2 mirrors the pattern at the end. I'll open the tracking issues once this one is in.

I've just rebased onto the latest kptdev/porch main, so the #981 copyright changes are picked up and CI is back to green.

Thanks again to everyone at the meeting.

liamfallon (dismissed):

pkg/cli/commands/rpkg/approve/command_test.go
```diff
@@ -104,14 +91,14 @@ func TestCmd(t *testing.T) {
 						RepositoryName: repoName,
 						ReadinessGates: []porchapi.ReadinessGate{
 							{
-								ConditionType: "kpt.dev.Specializer.specialize",
+								ConditionType: "nephio.org.Specializer.specialize",
```
mozesl-nokia: Is this actually a condition type from Nephio? If not, please change it back

pkg/cli/commands/rpkg/reject/command_test.go
```diff
@@ -226,3 +205,51 @@ func TestNewCommand(t *testing.T) {
...
+  name: t
+current-context: t
+users:
+- name: t
+`
+	if err := os.WriteFile(path, []byte(kubeconfigYAML), 0o600); err != nil {
+		t.Fatalf("write kubeconfig: %v", err)
+	}
+	return path
+}
```
mozesl-nokia: I think this function is duplicated a couple times

mozesl-nokia (dismissed):

pkg/cli/commands/rpkg/approve/command_test.go
thc1006: Thank you for catching this. The string was leftover test fixture data from before #981 swept the rest of the repository; it is not a real Nephio identifier, just a value used to make ReadinessGate.ConditionType and Condition.Type match each other so the readiness check returns false. Aligned to "kpt.dev.Specializer.specialize" so it matches what #981 settled on for the codebase. Fixed in commit f9f415d5.

pkg/cli/commands/rpkg/reject/command_test.go
thc1006: Good catch, thank you. Extracted writeTempKubeconfig as the exported rpkgutil.WriteTempKubeconfig in a new pkg/cli/commands/rpkg/util/testhelpers.go, matching the existing testhelpers_v1alpha2.go pattern. The three test files (pull, push, reject) now call the shared helper. Fixed in commit f9f415d5.

liamfallon (approved):

PR #1077, opened 2026-06-25
Limit Go-Git repo cache size and max in-memory file size

rendre-greyling:
## Description

**Problem:**
Go-Git creates a 96MB in-memory cache per repository by default and no limit on the maximum file size which can be stored in this cache. Every time commits are made and pushed, go-git creates memory objects for these and stores it in the cache. If you have hundreds of repositories and dozens to hundreds of package revisions per repository which are pushed, this memory consumption ramps up dramatically and remains in the system over time. Enabling draft & proposed pushing worsens this as it pushes more frequently.

**Solution:**
When a repository is opened in go-git, a new `NewObjectLRU`  is created which has a smaller maximum size. Added an additional option called `LargeObjectThreshold` which sets the maximum size for a file which for which a memory object can be created. If a file is larger than this threshold, it is not stored in the in-memory cache but rather as a file on the disk and streamed (the performance impact on this is negligible). **IMPORTANT: ensure your deployments have enough disk space**

Also made these options configurable through command line args. Defaults are:

- gogit-repo-cache-size=8MB
- gogit-cache-max-file-size=512KB

Sadly, when calling `InitEmptyRepository` a PlainInit is called (eventually in the call stack, it opens its own repository) which does not allow for the modification of these options. The WA is to reopen the repository with the same path and specifying the options. It's not pretty but works as intended. Go-git v6 is expected to enable this configuration and provide better large file handling, however it is still in alpha.

---

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Refactor
- [ ] Documentation
- [ ] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes  
- [x] Tests added/updated  

---

## AI Disclosure

- [ ] **I have used AI in the creation of this PR.**

### DIFF
--- a/pkg/cmd/server/start.go
+++ b/pkg/cmd/server/start.go
@@ -68,6 +68,9 @@ type PorchServerOptions struct {
 	DbMaxConnLifetime    time.Duration
 	DbPushDrafsToGit     bool
 
+	GoGitRepoCacheSize    int
+	GoGitCacheMaxFileSize int64
+
 	DefaultImagePrefix       string
 	FunctionRunnerAddress    string
 	LocalStandaloneDebugging bool // Enables local standalone running/debugging of the apiserver.
@@ -331,6 +334,8 @@ func (o *PorchServerOptions) Config() (*apiserver.Config, error) {
 				ExternalRepoOptions: externalrepotypes.ExternalRepoOptions{
 					LocalDirectory:         o.CacheDirectory,
 					UseUserDefinedCaBundle: o.UseUserDefinedCaBundle,
+					GoGitRepoCacheSize:     o.GoGitRepoCacheSize,
+					GoGitCacheMaxFileSize:  o.GoGitCacheMaxFileSize,
 				},
 				RepoOperationRetryAttempts: o.RepoOperationRetryAttempts,
 				CacheType:                  cachetypes.CacheType(o.CacheType),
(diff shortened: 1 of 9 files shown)


pkg/externalrepo/git/gogit.go
```diff
@@ -52,11 +54,14 @@ func initializeDefaultBranches(repo *git.Repository) error {
...
+func openRepository(path string, opts GitRepositoryOptions) (*git.Repository, error) {
 	dot := osfs.New(path)
-	storage := filesystem.NewStorageWithOptions(dot, cache.NewObjectLRUDefault(), filesystem.Options{
-		ExclusiveAccess: true,
-	})
+	objectCache := cache.NewObjectLRU(cache.FileSize(opts.GoGitRepoCacheSize) * cache.MiByte)
+	fileopts := filesystem.Options{
+		ExclusiveAccess:      true,
+		LargeObjectThreshold: opts.GoGitCacheMaxFileSize,
+	}
```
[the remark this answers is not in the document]
rendre-greyling: Done

pkg/cmd/server/start.go
```diff
@@ -331,6 +334,8 @@ func (o *PorchServerOptions) Config() (*apiserver.Config, error) {
 				ExternalRepoOptions: externalrepotypes.ExternalRepoOptions{
 					LocalDirectory:         o.CacheDirectory,
 					UseUserDefinedCaBundle: o.UseUserDefinedCaBundle,
+					GoGitRepoCacheSize:     o.GoGitRepoCacheSize,
```
efiacor, line 337: Do we need to be able to tune the repo ctr git cache independently?

https://github.com/kptdev/porch/blob/main/controllers/repositories/pkg/controllers/repository/cache.go#L149

rendre-greyling: It's more a of a nice to have than a necessity. I found 8MB to be a good default, but it's there so that if for example you have tens of thousands of repos, and relatively small PRs, you could drop it to 4MB and essentially halve your memory consumption

rendre-greyling: I see I completely misread this, I will add it as an option, but the Repo controller will already use the lower defaults in specified in gogit.go

liamfallon (dismissed):
This looks good to me. It allows tuning of go git but if tuning parameters are not specified it uses the default settings.

pkg/cmd/server/start.go
rendre-greyling: Added the options to the repo controller as well

Catalin-Stratulat-Ericsson (approved):

liamfallon (approved):

PR #1119, opened 2026-07-16
Move podevaluator to porch-server

dgyorgy-nokia:
# [Move podevaluator to porch-server]

---

## Description

- What changed: Moved podevaluator to porch-server from function-funner
- Why it’s needed: In a typical pipeline run where podevaluator is used, this eliminates an unnecessary gRPC exchange between porch-server and function-runner. The requests are sent directly from porch-server to the podexecutor, avoiding an additional gRPC round trip. The registration order of runners (builtin/executable/podevaluator) on porch-server is unchanged.
- How it works: 

---

## Related Issue(s)

- None #

---

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [x] Refactor
- [x] Documentation
- [ ] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes
- [x] Tests added/updated

---

## AI Disclosure

- [x] **I have used AI in the creation of this PR.**

If so, please describe how:
  - Cursor's composer 2.5 was used to review the changes.


api/generated/crds/config.porch.kpt.dev_functionconfigs.yaml
```diff
@@ -31,9 +31,6 @@ spec:
     - jsonPath: .status.apiServerObservedGeneration
       name: Server Applied
       type: integer
-    - jsonPath: .status.functionRunnerObservedGeneration
```
efiacor, line 34: Why are we dropping this? Not needed?

efiacor (commented):
We will need doc updates around this for sure as it's an arch change.

efiacor:
@dgyorgy-nokia rebase required

api/generated/crds/config.porch.kpt.dev_functionconfigs.yaml
dgyorgy-nokia: It's no longer needed as the FunctionConfig reconciler is removed from function-runner.
The porch-server decides whether an executable exists and only then asks the function-runner to run it.

CsatariGergely:
Would need some documentation. 
Could Dosu pick this up?

pkg/cmd/server/start.go
```diff
@@ -303,6 +309,12 @@ func (o *PorchServerOptions) Config() (*apiserver.Config, error) {
 		return nil, fmt.Errorf("error creating self-signed certificates: %w", err)
 	}
 
+	o.PodEvaluatorOptions.WrapperServerImage = os.Getenv(wrapperServerImageEnv)
+
+	if o.PodEvaluatorOptions.WrapperServerImage == "" {
```
efiacor, line 313: Will this cause the server to crash if not set? The error log infers that the pod evaluator is optional.

efiacor (commented):
Thanks for adding the documentation.
Could we add at least one crd e2e test that exercises the pod evaluator in the controller?

pkg/cmd/server/start.go
dgyorgy-nokia: Yes, it will crash.
I made a small adjustment to the log.

liamfallon:
Great work @dgyorgy-nokia , Could you rebase this one?

nagygergo:
Hey, is running 2 instances of `func/internal/podcachemanager.go` safe, in case it runs both in porch-controllers and porch-server?

liamfallon (approved):

kushnaidu (approved):

liamfallon:
We just need to address the outstanding comments and fix the CI to get this merged.

dgyorgy-nokia:
> Hey, is running 2 instances of `func/internal/podcachemanager.go` safe, in case it runs both in porch-controllers and porch-server?

Good point.
Unfortunately, this could cause problems.
The two podcachemanager instances aren't aware of each other and maintain their own separate state.
This could lead to a situation where one instance starts a pod and runs something on it, while the other instance tries to use a pod of the same type of krm function and may delete the existing one while it's still being used by the first instance.

One possible solution would be to give each of them a separate namespace, which would prevent this conflict. However, if both podcachemanager instances receive the same type of request, they could end up creating duplicate pods at the same time, which would be wasteful.

The other option would be to centralize the pod tracking and have both instances use a shared pool. This would be a much larger change, though, and there is already a proposal/study for this approach, which also includes the move of podevaluator to kpt, in which case this change would no longer be necessary.

efiacor:
@dgyorgy-nokia Kpt tech meeting decision:
2026-09-23: Plan is to document that ONLY 1 pod evaluator should run in the overall porch deployment.
 - Update PR to define only 1 running instance in our dev/script setups (preferably porch server?)

PR #1122, opened 2026-07-18
Refactor Performance Tests and expand Monitoring Stack

rendre-greyling:
## Refactor Performance Tests and expand Monitoring Stack

This PR is a continuation of [https://github.com/kptdev/porch/pull/561](https://github.com/kptdev/porch/pull/561) which has been closed due to being out of date.

**NOTE: This PR is ~30K lines - a large majority of this is generic test kpt packages (~19k lines) and large Grafana json dashboards**

### New Performance Tests and Suite

The previous performance tests were split between 2 performance packages, one under `test/performance` folder, the other under `test/e2e/performance` which was the Iterative test. These tests were outdated and lacked certain metrics. Features of both have now been consolidated under one `test/performance` folder. 

The new Performance Tests work with both the `v1alpha1` and `v1alpha2` api's of porch and can be expanded in future.

There are now 2 new tests which replace the previous tests and functionality:

- **Load Test:** Creates a preset amount of repositories, packages, and revisions to simulate load on the Porch system.
- **Maximum Package Revisions:** Tests the maximum number of package revisions that can be handled by Porch in a single repository.

See the `test/performance/README.md` for all info on the data that is collected by the tests and customization arguments. The data is stored in csv files for later analysis. The data is also exposed to OTel/Prometheus if the `-enable-prometheus` is set.

The Package Revisions which are deployed can be selected from 3 generic Kpt packages:
- small-package: Minimal footprint and 2 basic kpt mutators for 1 file in the package.
- complex-package: Multiple mutators and validators across a dozen files in the package to simulate loads on function pods.
- large-package: Multiple mutators and validators across a couple of files in the package and one very large (~17k lines) `deployment.yaml` file to simulate the impact of large files on the system.

### New Performance Monitoring Stack

The OTel interface has been expanded to expose various new API metrics as well as Memory and CPU data. A new `deploy-monitoring.sh` script along with `make` targets has been added which will deploy Grafana, Prometheus, Postgres Exporter (for the porch DB cache metrics), Pyroscope (using alloy to get the data from pprof) and Jaeger pods in the kind cluster for visual analysis of the data.

The make targets are:
- `make deploy-monitoring`: Deploys the base Grafana, Prometheus and Postgres Exporter pods, services and sets up port forwarding.
- `make deploy-monitoring-jaeger`: Deploys the Jaeger pod, service and sets up port forwarding.
- `make deploy-monitoring-pyroscope`: Deploys the Pyroscope and Alloy pods, services and port forwarding.
- `make cleanup-monitoring`: Deletes all monitoring pods, services and port forwarding.
- `make restart-monitoring`: Restarts all monitoring pods, services and port forwarding.

Grafana has 4 dashboards:

- Porch API: PR, PRR and external Git requests time taken per API call, as well as req/s per user account. (these metrics are tracked for both `v1alpha1` and `v1alpha2`)
- Porch Performance Test Dashboard: All recorded metrics from the performance tests
- Porch Resource Usage: Memory and CPU usage of the porch-server, function runner, controllers, DB Cache and function pods.
- Porch Postgres Dashboard: Displays metrics of the Postgres DB such as QPS, deadlocks and cache hits and misses.
- Porch Package Size Dashboard: remains unchanged from: [https://github.com/kptdev/porch/pull/1018](https://github.com/kptdev/porch/pull/1018) except for smaller vertical height of the panel.
- Pyroscope – Porch profiling: Advanced CPU, Memory and Go profiling data with the use of flame graphs.

Prometheus UI can be accessed by [http://localhost:9092/](http://localhost:9092/)

Pyroscope UI can be accessed by [http://localhost:4040/](http://localhost:4040/) after running the make deploy-monitoring-pyroscope` for more granularity.

Jaeger UI can be accessed by [http://localhost:16686/](http://localhost:16686/) after running the make deploy-monitoring-jaeger` for advanced tracing.

---

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Enhancement
- [x] Refactor
- [ ] Documentation
- [x] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes  
- [x] Tests added/updated  

---

## Testing Instructions (Optional)
1. Setup kind cluster
2. Follow instructions in `test/performance/README.md`

---

## AI Disclosure
[x] I have used AI in the creation of this PR.

- Cursor composer 2.5 and Sonnet 4.6 used for the creation of the generic kpt packages and Grafana dashboards, squaring away the scripts and deployments as well as the csv creation. Also assisted in splitting the original implementation of the performance tests to be able to use both API types.


pkg/registry/porch/packagerevision_approval.go
```diff
@@ -53,7 +57,13 @@ func (a *packageRevisionApproval) NamespaceScoped() bool {
...
 func (a *packageRevisionApproval) Get(ctx context.Context, name string, _ *metav1.GetOptions) (runtime.Object, error) {
 	ctx, span := tracer.Start(ctx, "[START]::packageRevisionApproval::Get", trace.WithAttributes())
-	defer span.End()
+	start := time.Now()
+	defer func() {
+		span.End()
+		telemetry.RecordAPICallDuration(praTelemetryName, "GET", telemetry.APIVersionV1Alpha1, time.Since(start).Seconds())
+	}()
+
+	telemetry.RecordRequestCount(ctx, prTelemetryName, "GET", telemetry.APIVersionV1Alpha1)
```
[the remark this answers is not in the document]
rendre-greyling: fixed

pkg/registry/porch/packagerevision_approval.go
```diff
@@ -70,7 +80,13 @@ func (a *packageRevisionApproval) Get(ctx context.Context, name string, _ *metav
...
 	updateValidation rest.ValidateObjectUpdateFunc, _ bool, _ *metav1.UpdateOptions) (runtime.Object, bool, error) {
 	ctx, span := tracer.Start(ctx, "[START]::packageRevisionApproval::Update", trace.WithAttributes())
-	defer span.End()
+	start := time.Now()
+	defer func() {
+		span.End()
+		telemetry.RecordAPICallDuration(praTelemetryName, "UPDATE", telemetry.APIVersionV1Alpha1, time.Since(start).Seconds())
+	}()
+
+	telemetry.RecordRequestCount(ctx, prTelemetryName, "UPDATE", telemetry.APIVersionV1Alpha1)
```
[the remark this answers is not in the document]
rendre-greyling: fixed

test/performance/logger.go
```diff
@@ -19,49 +19,209 @@ import (
...
 	"os"
 	"path/filepath"
-	"testing"
+	"sync"
 	"time"
+
+	pkgerrors "github.com/pkg/errors"
 )
 
+const timeDateFormat = "2006-01-02 15:04:05"
```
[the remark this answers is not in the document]
rendre-greyling: fixed

internal/telemetry/pprof.go
```diff
@@ -0,0 +1,92 @@
...
+			klog.Errorf("Error shutting down profiling server: %v", err)
+		}
+	}
+}
+
+func (p *Profiling) serve() {
+	klog.Infof("Starting profiling server on port :%d", p.port)
+	if err := p.server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		klog.Errorf("Error starting metrics server: %v", err)
+	}
```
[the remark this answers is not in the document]
rendre-greyling: fixed

deployments/porch/3-porch-server.yaml
```diff
@@ -143,5 +160,9 @@ spec:
       protocol: TCP
       targetPort: 9464
       name: metrics
+    - port: 8080
+      protocol: TCP
+      targetPort: 8080
+      name: pprof
```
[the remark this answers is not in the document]
rendre-greyling: fixed - was unnecessary

deployments/porch/2-function-runner.yaml
```diff
@@ -118,3 +134,7 @@ spec:
       protocol: TCP
       targetPort: 9464
       name: metrics
+    - port: 8080
+      protocol: TCP
+      targetPort: 8080
+      name: pprof
```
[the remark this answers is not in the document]
rendre-greyling: fixed - was unnecessary

deployments/porch/9-controllers.yaml
```diff
@@ -125,5 +146,9 @@ spec:
       protocol: TCP
       targetPort: 9464
       name: metrics
+    - port: 8080
+      protocol: TCP
+      targetPort: 8080
+      name: pprof
```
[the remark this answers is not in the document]
rendre-greyling: fixed - was unnecessary

.env.template
```diff
@@ -1,3 +1,4 @@
 # PORCH_GHCR_PREFIX_URL=<custom registry>/kptdev/krm-functions-catalog
 # DOCKERHUB_MIRROR=<mirror registry url>
 # DB_MAVEN_MIRROR=<mirror repository url>
+# GRAFANA_ADMIN_PW<password for grafana UI>
```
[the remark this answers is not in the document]
rendre-greyling: fixed

scripts/deploy-monitoring.sh
```diff
@@ -70,9 +112,32 @@ check_kpt() {
...
 
+is_base_deployed() {
+    local deployment
+    for deployment in prometheus grafana postgres-exporter; do
+        if ! kubectl get deployment "$deployment" -n "$NAMESPACE" &> /dev/null; then
+            return 1
+        fi
+    done
+    return 0
+}
```
JamesMcDermott, line 123: Now that the script has multiple commands, isn't it likely that the `deploy` option could be run multiple times after the initial deployment? That isn't quite idempotent for Grafana - it'll generate a new password and patch it into `secret/grafana-admin-creds`, meaning the original Grafana password will be lost and the new one won't work.

Perhaps the `deploy` flow could do one of these?
- get the current password value from `grafana-admin-creds` before `prepare_manifests` is called so the secret is patched with itself
  - could be part of this `is_base_deployed` function? would need adding an `is_base_deployed` call to the `deploy` flow
- restart the Grafana pod to take the new password from the secret, and re-do the port-forward

deployments/metrics/pyroscope-deployment.yaml
```diff
@@ -0,0 +1,144 @@
+# Copyright 2026 The Nephio Authors
```
JamesMcDermott:
```suggestion
# Copyright 2026 The kpt Authors
```
Also applies to other files - quick search found `pprof.go` and `pyroscope.go`; may be others with mentions of Nephio.

internal/telemetry/pprof.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2025 The Nephio Authors
```
JamesMcDermott:
```suggestion
// Copyright 2026 The kpt Authors
```

test/performance/performance_suite.go
```diff
@@ -0,0 +1,667 @@
...
+	namespace          = flag.String("namespace", "porch-metrics", "Kubernetes namespace to use for the test")
+	porchAPIVersion    = flag.String("api-version", string(PorchAPIV1Alpha1), "Porch API version to test (v1alpha1 or v1alpha2)")
+	numRepos           = flag.Int("repos", 1, "Number of repositories to create")
+	numPackages        = flag.Int("packages", 1, "Number of packages per repository")
+	numRevisions       = flag.Int("revisions", 1, "Number of package revisions per package")
+	repoParallelism    = flag.Int("repo-parallelism", 1, "Number of repositories to create in parallel")
+	packageParallelism = flag.Int("package-parallelism", 1, "Number of packages to create in parallel per repository")
+	errorRate          = flag.Float64("error-rate", 0.1, "Maximum percentage of package revisions allowed to fail lifecycle transition")
+	enableDeletion     = flag.Bool("enable-deletion", false, "Enable deletion of package revisions at the end of the test")
+	enablePrometheus   = flag.Bool("enable-prometheus", false, "Enable Prometheus metrics server on port 9091")
```
JamesMcDermott: Wrong port:
```suggestion
	enablePrometheus   = flag.Bool("enable-prometheus", false, "Enable Prometheus metrics server on port 9095")
```

test/performance/performance_suite.go
```diff
@@ -0,0 +1,667 @@
...
+func getEnvWithDefault(key, defaultValue string) string {
+	_ = godotenv.Load(filepath.Join("..", "..", ".env"))
+	if value := os.Getenv(key); value != "" {
+		return value
+	}
+	return defaultValue
+}
+
+func (t *PerfTestSuite) SetupSuite() {
+	if os.Getenv("LOAD_TEST") != "1" && os.Getenv("MAX_PR_TEST=1") != "1" {
```
JamesMcDermott: This is referring to the variable named `MAX_PR_TEST=1` rather than `MAX_PR_TEST` as used in `porch_metrics_test.go` - presume it should be unified to `MAX_PR_TEST`?

To prevent things like this in future, you could yoke it all to a single field on the `PerfTestSuite` type, like I did with the DB_CACHE variable in the e2e tests - see here: https://github.com/kptdev/porch/commit/0b526cad79e076d948c5da0101790f5fe3f10101#diff-b0086a4d69df5526858b2bae5da63347d8411bab75abc25afbb87798de8fed38R146. That'd have one check of the environment variable in SetupSuite (or earlier) and you'd be able to refer to the field everywhere after that.

internal/telemetry/metrics.go
```diff
@@ -16,24 +16,70 @@ package telemetry
...
+
+	// Performance tests related vars
+	perfOperationDuration           metric.Float64Histogram
+	perfOperationCounter            metric.Float64Counter
+	perfRepositoryCounter           metric.Float64Counter
+	perfPackageCounter              metric.Float64Counter
+	perfPackageRevisionCounter      metric.Float64Counter
+	perfLifecycleTransitionDuration metric.Float64Histogram
+	perfTestRunInfoGauge            metric.Float64Gauge
+	perfActiveOperations            metric.Float64UpDownCounter
```
JamesMcDermott: If these are only used in the performance tests, could they be more closely associated with those, instead of being baked into all the production code? It looks like they're only called from stuff under the `test/performance` path - would it be possible to refactor them out to there?

internal/telemetry/metrics.go
```diff
@@ -16,24 +16,70 @@ package telemetry
...
+		"porch_api_call_duration_seconds",
+		metric.WithDescription("Duration of porch API calls in seconds."),
+		metric.WithUnit("s"),
+		metric.WithExplicitBucketBoundaries(
+			0.001, 0.002, 0.004, 0.008, 0.016, 0.032, 0.064, 0.128,
+			0.256, 0.512, 1.024, 2.048, 4.096, 8.192, 16.384, 32.768,
+		),
+	)
+	if err != nil {
+		panic(fmt.Sprintf("failed to create porch_api_call_duration_seconds: %v", err))
```
JamesMcDermott: Do we really need to panic and take down the entire process if a metric fails to create, or could this just log and return the error instead as the `prResourcesSize*` ones do?
(same for all metrics introduced here)
(if possible to refactor the performance-test metrics out, we could keep the panics for those in their new location, but it seems even that could better serve by Fatal'ing the test)

deployments/porch/3-porch-server.yaml
```diff
@@ -54,7 +54,7 @@ spec:
               memory: 256Mi
               cpu: 250m
             limits:
-              memory: 512Mi
+              memory: 2Gi
```
JamesMcDermott: Why are we quadrupling porch-server's default memory limit?
(do the performance tests need it? could the performance suite setup try and patch it in instead of making it the default?)

internal/telemetry/metrics.go
```diff
@@ -16,24 +16,70 @@ package telemetry
...
 	"go.opentelemetry.io/otel/metric"
+	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/klog/v2"
 )
 
 const meterName = "github.com/kptdev/porch"
 
+const (
+	APIVersionV1Alpha1 = "v1alpha1"
+	APIVersionV1Alpha2 = "v1alpha2"
```
JamesMcDermott: (not a deal-breaker, but maybe:)
```suggestion
	APIVersionV1Alpha1 = v1alpha1.SchemeGroupVersion.Version
	APIVersionV1Alpha2 = v1alpha2.SchemeGroupVersion.Version
```
(import "github.com/kptdev/porch/api/porch/v1alpha1" and "github.com/kptdev/porch/api/porch/v1alpha2")
(and they'd have to be moved to the var section)

internal/telemetry/pyroscope.go
```diff
@@ -0,0 +1,103 @@
...
+	PyroscopeServerEnvVar      = "PYROSCOPE_SERVER"
+	PyroscopeAppNameEnvVar     = "PYROSCOPE_APP_NAME"
+	PyroscopeAuthUserVar       = "PYROSCOPE_AUTH_USER"
+	PyroscopeAuthPassVar       = "PYROSCOPE_AUTH_PASSWORD" // #nosec G101 -- env var name, not a credential (nolint:gosec)
+	PyroscopeLogsEnabledEnvVar = "PYROSCOPE_LOGS_ENABLED"
+)
+
+const defaultPyroscopeAppName = "porch-server"
+
+type PyroscopeProfiling struct {
```
JamesMcDermott: The PyroscopeProfiling setup doesn't appear to be used anywhere - is there some wiring left to do (or is Grafana/Pyroscope capable of picking up on it automatically in some way?)?

internal/telemetry/pprof.go
```diff
@@ -0,0 +1,92 @@
...
+			return
+		}
+		p.port = parsed
+	} else {
+		return
+	}
+
+	// Enable profiling for mutex and block profiles
+	runtime.SetMutexProfileFraction(1)
+	runtime.SetBlockProfileRate(1)
```
JamesMcDermott, line 51: Kiro (AI review) suggests this will result in dumping every mutex and blocking event, resulting in additional CPU overhead and performance regression... thoughts? If accurate, would it be possible to settle on reasonable higher values or yoke these to an install-time flag so full profiling can be kept for debugging or testing?

internal/telemetry/pyroscope.go
```diff
@@ -0,0 +1,103 @@
+// Copyright 2026 The Nephio Authors
```
JamesMcDermott:
```suggestion
// Copyright 2026 The kpt Authors
```

pkg/registry/porch/packagerevision_approval.go
JamesMcDermott:
```suggestion
// Copyright 2022, 2026 The kpt Authors
```

test/performance/logger.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024, 2026 The kpt Authors
+// Copyright 2026 The kpt Authors
```
JamesMcDermott: The file was added over the 2024/2025 boundary... consider:
```suggestion
// Copyright 2025-2026 The kpt Authors
```

test/performance/porch_metrics_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024, 2026 The kpt Authors
+// Copyright 2026 The kpt Authors
```
JamesMcDermott, line 1: The file was added over the 2024/2025 boundary... consider:
```suggestion
// Copyright 2025-2026 The kpt Authors
```

test/performance/types.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024, 2026 The kpt Authors
+// Copyright 2026 The kpt Authors
```
JamesMcDermott: The file was added over the 2024/2025 boundary... consider:
```suggestion
// Copyright 2025-2026 The kpt Authors
```

test/performance/csv_generation.go
```diff
@@ -0,0 +1,335 @@
+// Copyright 2026 The Kpt Authors
```
JamesMcDermott:
```suggestion
// Copyright 2026 The kpt Authors
```

internal/telemetry/metrics.go
```diff
@@ -16,24 +16,70 @@ package telemetry
...
+	ControllerUser     = "packagerevision-controller"
+
+	ResourcePackageRevision          = "PackageRevision"
+	ResourcePackageRevisionResources = "PackageRevisionResources"
+	ResourceExternalRepo             = "ExternalRepo"
+)
+
 var (
+	apiCallDurationSeconds  metric.Float64Histogram
+	RequestsTotal           metric.Float64Counter
```
JamesMcDermott: Should `RequestsTotal` be exported? It seems a bit out of place when all the others aren't exported.

```suggestion
	requestsTotal           metric.Float64Counter
```

func/server/server.go
```diff
@@ -244,6 +245,9 @@ func buildFnConfigReconciler(o *options, scheme *runtime.Scheme) (*reconciler.Fu
 	mgr, err := ctrl.NewManager(restCfg, ctrl.Options{
 		Scheme: scheme,
 		Cache:  cacheOpts,
+		Metrics: metricsserver.Options{
+			BindAddress: "0",
+		},
```
JamesMcDermott, line 252: Could you add a comment explaining why we're setting this?
(according to Kiro it disables the built-in metrics endpoint?)

pkg/apiserver/config.go
```diff
@@ -295,6 +296,9 @@ func (c completedConfig) buildFunctionConfigReconciler(ctx context.Context, sche
 	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme: scheme,
 		Cache:  cacheOpts,
+		Metrics: metricsserver.Options{
+			BindAddress: "0",
+		},
```
JamesMcDermott, line 284: Could you add a comment explaining why we're setting this?
(according to Kiro it disables the built-in metrics endpoint?)

internal/telemetry/metrics.go
```diff
@@ -55,10 +101,154 @@ func InitMetrics() (err error) {
...
+	recordExternalRepoDuration(op, time.Since(start).Seconds())
+	RecordExternalRepoRequestCount(ctx, op)
+}
+
+func recordExternalRepoDuration(op string, durationSeconds float64) {
+	if apiCallDurationSeconds == nil {
+		return
+	}
+	apiCallDurationSeconds.Record(context.Background(), durationSeconds,
+		metric.WithAttributes(
```
JamesMcDermott, line 164: Should this also include an "api_version" attribute, even with a dummy ("N/A"?) value, to align with the attributes recorded to the same metric by `RecordAPICallDuration`?

test/performance/driver.go
```diff
@@ -0,0 +1,349 @@
...
+	duration := time.Since(start)
+
+	t.recordPkgRevMetric(repoName, pkgName, revisionNum, pkgRevCreate, OperationMetrics{
+		Operation: fmt.Sprintf("%s:%d", pkgRevCreate, revisionNum),
+		Duration:  duration,
+		Error:     err,
+		Timestamp: start,
+	})
+	t.recordPerfMetric(pkgRevCreate, repoName, pkgName, duration, err)
+	if t.enablePrometheus {
```
JamesMcDermott: There's lots of `t.enablePrometheus` checks in this file especially - could fold them all into an extracted method like `t.recordPerfMetric()`.

internal/telemetry/metrics.go
```diff
@@ -55,10 +101,154 @@ func InitMetrics() (err error) {
...
+
+func RecordExternalRepoRequestCount(ctx context.Context, op string) {
+	if RequestsTotal == nil {
+		return
+	}
+	RequestsTotal.Add(context.Background(), 1,
+		metric.WithAttributes(
+			attribute.String("resource", ResourceExternalRepo),
+			attribute.String("op", op),
+			attribute.String("user", getK8sUserName(ctx)),
```
JamesMcDermott: Assuming this is going to be called in Git operations, are we guaranteed to have a K8sUserName populated (often enough for it to be meaningful)? Is there maybe a Git user that could be obtained to provide more context-relevant information?

JamesMcDermott (commented):
- [x] **I have used AI in the creation of this ~~PR~~ review.**
  - Amazon Q/Kiro to highlight reviewable points
    - wording and suggestions mostly my own :)

internal/telemetry/pyroscope.go
rendre-greyling: Oh thanks for catching this. Pyroscope.go should have been removed when I brought over the changes from the old MR. All metrics for it is being gathered by pprof already.

test/performance/README.md
Catalin-Stratulat-Ericsson, line 1: im sorry to be that guy. but i think the contents of this should be in the documentation rather than a nestled README file.

rendre-greyling: I agree, but I want to do this in a separate PR.  Imo it will need to be referenced under opentelemerty, local-dev-deployment, a new testing page and probably under development-environment under code-contribution. This PR has gotten quite unwieldy, so it will be easier to review this all in a new PR.

Catalin-Stratulat-Ericsson: i agree with you. its already a massive PR even without the package. im just worried about we merge this PR in and then forget to add any docs to it later as its not priority

rendre-greyling: I'll create an issue for it and assign it to me. I think I'll do the push drafts to git first since that seems to be much higher prio atm.

Catalin-Stratulat-Ericsson: just have a look at this PR and the others linked in it for the stuff i fixed on it quick before so were not duplicating :)
https://github.com/kptdev/porch/pull/1143

internal/telemetry/pprof.go
rendre-greyling: Yes, Kiro is accurate here that it does cause additional CPU overhead, but from my testing this is negligible on performance (I've tested this with a few thousand PRs). Also, pprof is only really used to expose these metrics to Pyroscope via Alloy. It should only ever really be enabled for debugging and I think that it's valuable to have these set so aggressive. What I will do instead is have the deployment script patch the envvar for the PORCH_PPROF_PORT in when deploying Pyroscope. If no port is set, then pprof is disabled

internal/telemetry/metrics.go
rendre-greyling: I don't really think this is necessary. The external repo metrics are specifically for git which is pretty much shared between the 2 APIs. If we expand in future on the metrics from git and on the controller metrics, then maybe we can revisit it?

internal/telemetry/metrics.go
rendre-greyling: No, it's not guaranteed. On FETCH it will most likely display as <UNKOWN>. Much more likely to have a k8s Username on PUSH.

func/server/server.go
rendre-greyling: I left a comment. It does disable the built-in metrics endpoint which is 8080 by default (which pprof also uses). I can enable it and have pprof use a different port, but it would be redundant to enable this because these same metrics are already exposed via Otel. Thoughts on this are welcome :)

pkg/apiserver/config.go
rendre-greyling: See above comment

scripts/deploy-monitoring.sh
rendre-greyling: Decided to take option 1. Just going to leave a note here that .env.template has a new ennvar: GRAFANA_ADMIN_PW=<password for grafana UI>

You can set it if you don't want to have a randomly generated password during development.

deployments/porch/3-porch-server.yaml
rendre-greyling: Accidentally left in during testing of large amounts of large PRs. Reverted to 512Mi. User should set according to their needs and scale of tests being performed.

test/performance/driver.go
rendre-greyling: Extracted, but into multiple functions due to differing conditions

internal/telemetry/metrics.go
rendre-greyling: Also, I prefer the k8s username because I think there will be more cases where this is useful on push instead of a Git User.

func/server/server.go
JamesMcDermott: Yep, comment looks good - you're right, no need for redundant exposure if we're dedicated to OTel anyway. I don't have a strong opinion on the matter, just it's good to have an explanation to help prevent it being rashly reverted in the future.

pkg/apiserver/config.go
JamesMcDermott: Comment looks good. 👍

scripts/deploy-monitoring.sh
JamesMcDermott: Ah yes, I see it now. 👍
(is it wise to leave it like that in case someone decides to leave it lying around in a production environment?)

test/performance/driver.go
JamesMcDermott: Yep, multiple functions are fine.
OK, so if I'm reading `defer t.trackPerfActiveOperation(pkgRevCreate)()` correctly, most of `trackPerfActiveOperation`, including the `RecordPerfActiveOperation` with delta `1`, should run immediately, and the only thing that's actually deferred is the `RecordPerfActiveOperation` with delta `-1`, correct?
(nice!😀)

rendre-greyling: Yep exactly

scripts/deploy-monitoring.sh
rendre-greyling: Well, my thoughts are that the metrics deployments and the script are more for local deployments, testing, analysis and debug. The dashboards can of course be used for production as well as the deployments as a base template. I would certainly hope though that the script is not used as is for production and that a more robust solution is used. Also, at the very least an ingress and not a port forward. More likely as well is that porch will not be the only thing running in prod, and so the Grafana instance might have many other dashboards and features set up.

I would honestly prefer not even having a password set (this is more a me being lazy problem :) ) but I think the envvar is a decent middle ground.

JamesMcDermott: Fair enough; it was more on the line of having it in the .env.template as an example - but I take your point, and of course it's not enabled by default.

JamesMcDermott (approved):

internal/telemetry/metrics.go
```diff
@@ -16,24 +16,64 @@ package telemetry
...
 	m := otel.Meter(meterName)
 
+	apiCallDurationSeconds, err = m.Float64Histogram(
+		"porch_api_call_duration_seconds",
+		metric.WithDescription("Duration of porch API calls in seconds."),
+		metric.WithUnit("s"),
+		metric.WithExplicitBucketBoundaries(
+			0.001, 0.002, 0.004, 0.008, 0.016, 0.032, 0.064, 0.128,
+			0.256, 0.512, 1.024, 2.048, 4.096, 8.192, 16.384, 32.768,
+		),
```
mozesl-nokia: I don't really like these numbers being inline... you could generate these in a for loop with 2 inputs maybe? You can keep them consts for now, something like "apiCallDurationStartingBucket" and "...BucketCount"

internal/telemetry/metrics.go
```diff
@@ -92,3 +212,10 @@ func RecordPackageRevisionResourcesSize(ctx context.Context, prKey repository.Pa
 	}
 	prResourceSizeGauge.Record(ctx, resourcesSize, metric.WithAttributeSet(attributes))
 }
+
+func getK8sUserName(ctx context.Context) string {
+	if user, ok := request.UserFrom(ctx); ok {
+		return user.GetName()
+	}
+	return "<UNKNOWN>"
+}
```
mozesl-nokia: Why has this been copied here? Import cycle?

internal/telemetry/otel.go
```diff
@@ -116,6 +120,10 @@ func SetupOpenTelemetry(ctx context.Context) (*OTelResources, error) {
 		return nil, err
 	}
 
+	prof := &Profiling{}
+	prof.Start()
+	res.profiling = prof
```
mozesl-nokia:
```suggestion
	res.profiling = &Profiling{}
	res.profiling.Start()
```

internal/telemetry/pprof.go
```diff
@@ -0,0 +1,92 @@
...
+	mux.HandleFunc("/debug/pprof/cmdline", httppprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", httppprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", httppprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", httppprof.Trace)
+	mux.Handle("/debug/pprof/heap", httppprof.Handler("heap"))
+	mux.Handle("/debug/pprof/goroutine", httppprof.Handler("goroutine"))
+	mux.Handle("/debug/pprof/threadcreate", httppprof.Handler("threadcreate"))
+	mux.Handle("/debug/pprof/block", httppprof.Handler("block"))
+	mux.Handle("/debug/pprof/mutex", httppprof.Handler("mutex"))
+	mux.Handle("/debug/pprof/allocs", httppprof.Handler("allocs"))
```
mozesl-nokia, line 65: Would it not be better to just set up a redirect to the default pprof endpoints instead of manually re-handling every endpoint? Or can you not start pprof on a custom prefix?

go.mod
```diff
@@ -70,7 +71,10 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
-require sigs.k8s.io/cli-utils v0.37.2 // indirect
+require (
+	github.com/grafana/pyroscope-go/godeltaprof v0.1.11 // indirect
+	sigs.k8s.io/cli-utils v0.37.2 // indirect
+)
```
mozesl-nokia: Could you fix these being in a separate block?

test/performance/performance_suite.go
mozesl-nokia, line 1: Also, you could should replace the t.T().Fatalf calls with t.Require().NoError or such

internal/telemetry/pprof.go
rendre-greyling: I believe we can't redirect the pprof URL's here because we're using a custom mux. We would need to use DefaultServeMux. At least that's how I understand it

mozesl-nokia (approved):

PR #1146, opened 2026-08-05
Redesign DB Draft and Proposed Pushing

rendre-greyling:
## Description
The previous implementation of pushing draft and proposed package revisions mimicked the CR cache by pushing a draft and proposed package revision after every update. This resulted in a Package Revision being pushed at minimum 4 times from creation to Published state. Many further issues were present such as race conditions, unwanted deletions of Package Revisions and increased load on the external Git instance. The design was quite terrible and unstable - I can say this since I implemented the feature :)

This new implementation completely removes this logic and moves the push logic for draft and proposed Package Revisions to the sync process in the DB cache (which is triggered by the Repository controller). 

- When a Package Revision is created, it is not pushed to the external git. If it is subsequently updated, proposed and approved before the next sync runs, then only the approved tag for the approved Package Revision will be present in the external git (the same as without using the push drafts to git feature).
- If a Package Revision is created, but not approved and the next sync runs, the Package Revision will be pushed to the external git as a branch.
- If said Package Revision is Proposed and the next sync runs, the Package Revision is pushed to the external git as a proposed branch (the draft branch is deleted). If then approved, the proposed branch is deleted, and the approved tag is created.
- A proposed branch does not require there to have been a draft branch before it and an approval does not require there to have been a draft or proposed branch. It will delete either one which is present when creating the tag.

Reasoning of this change:

- Preserves the ability to only have to back up the external git data as draft and proposed Package Revisions will be reflected in the external git and the DB Cache can be rebuilt from this.
- Significantly decreased load to the external Git server during automated hydration operations since in most cases a Package Revision will be created and approved in a short span of time before a sync runs. 
- Increased speed of Hydration, less pushes = less hydration time and less load on external git server.
- It has been observed in some cases that deleting draft and proposed branches can be expensive and take significant amount of time in the external git, so eliminating this is crucial.

---

## Related Issue(s)
Related to: https://github.com/kptdev/porch/pull/1143

---

## Usage
Set the `db-push-drafts-to-git` flag in the repository controller and porch-server to true.

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Enhancement
- [x] Refactor
- [ ] Documentation
- [x] Tests
- [ ] Other: ________

---

## Checklist
- [x] Self-reviewed changes
- [x] Tests added/updated

---

## AI Disclosure

- [x] **I have used AI in the creation of this PR.**

composer 2.5 for UT generation and assistance with e2e tests.


pkg/cache/dbcache/util.go
```diff
@@ -44,3 +50,109 @@ func setValueFromJSON(jsonValue string, value any) {
 		klog.Errorf("unmarshal of json value %v failed, %v ", jsonValue, err)
 	}
 }
```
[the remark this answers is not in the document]
rendre-greyling: resolved

pkg/cache/dbcache/util.go
```diff
@@ -44,3 +50,109 @@ func setValueFromJSON(jsonValue string, value any) {
...
+	return globalLockManager.getLock(repoKey.String())
+}
+
+func getOrInsertPkgLock(pkgKey repository.PackageKey) *sync.Mutex {
+	return globalLockManager.getLock(pkgKey.String())
+}
+
+func deletePkgLock(pkgKey repository.PackageKey) {
+	globalLockManager.deleteLock(pkgKey.String())
+}
```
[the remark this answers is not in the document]
rendre-greyling: resolved

pkg/cache/dbcache/dbrepository.go
```diff
@@ -556,9 +482,15 @@ func (r *dbRepository) Refresh(ctx context.Context) error {
...
 
+	repoMutex := getOrInsertRepoLock(r.Key())
+	repoMutex.Lock()
+	defer repoMutex.Unlock()
+
+	r.repositorySync.mutex.Lock()
 	if err := r.externalRepo.Refresh(ctx); err != nil {
 		return err
 	}
+	r.repositorySync.mutex.Unlock()
```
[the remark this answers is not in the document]
rendre-greyling: resolved, but not with defer, SyncOnce will also try to lock the mutex

test/e2e/suiteutils/gitea_test_utils.go
```diff
@@ -126,3 +143,283 @@ func RecreateGiteaRepo(t *testing.T, repoName string) {
...
+}
+
+func (t *TestSuite) CreateGiteaRepoNoCleanup(repoName string) string {
+	t.T().Helper()
+
+	body := fmt.Sprintf(`{"name":%q,"auto_init":true,"readme":"Default"}`, repoName)
+	req, _ := http.NewRequest("POST", t.GetGiteaApiURL()+"/api/v1/user/repos", strings.NewReader(body))
+	req.SetBasicAuth(t.GiteaUser, t.GiteaPassword)
+	req.Header.Set("Content-Type", "application/json")
+
```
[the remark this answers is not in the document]
rendre-greyling: resolved

pkg/cache/dbcache/dbreposync_test.go
```diff
@@ -263,14 +263,17 @@ func (t *DbTestSuite) TestDBRepoSyncWithPushDraftsToGit_DraftOnlyInCacheRemoved(
...
 	t.Require().NoError(err)
 
-	// Do not add the draft to the external repo. Sync should treat it as "cached only" and remove it.
+	// Do not add the draft to the external repo. Sync should queue a git push instead of deleting it.
 	// Explicitly trigger sync
 	err = testRepo.repositorySync.SyncOnce(ctx)
 	t.Require().NoError(err)
 
+	// Allow the async PushDraftPackageRevision goroutine to finish.
+	time.Sleep(200 * time.Millisecond)
```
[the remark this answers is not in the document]
rendre-greyling: resolved

pkg/cache/dbcache/dbreposync.go
```diff
@@ -250,26 +243,132 @@ func (s *repositorySync) cacheExternalPRs(ctx context.Context, externalPrMap map
...
+			klog.Infof("repositorySync %+v: reconcile %+v: DB changed since last push, pushing to git", s.repo.Key(), prKey)
+			s.enqueuePush(ctx, cachedPR)
+		}
+	}
+}
+
+func (s *repositorySync) enqueuePush(ctx context.Context, pr *dbPackageRevision) {
+	pushCtx := context.WithoutCancel(ctx)
+	go PushDraftPackageRevision(pushCtx, s.repo.Key(), pr)
+}
```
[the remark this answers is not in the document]
rendre-greyling: Whilst this is true, the volume expected for these pushes is very low. Fix would be to put this in a worker pool and limit the number of parallel pushes introducing more cmd args.

I would prefer to hold off on this now as there is a common worker pool implementation in the works and would rather use that.

pkg/cache/dbcache/dbpushtogit.go
```diff
@@ -0,0 +1,240 @@
+// Copyright 2025 The kpt Authors
```
efiacor:
```suggestion
// Copyright 2026 The kpt Authors
```

deployments/porch/9-controllers.yaml
```diff
@@ -51,6 +51,7 @@ spec:
         imagePullPolicy: IfNotPresent
         args:
         - --repositories.cache-type=DB
+        - --repositories.push-drafts-to-git=false
```
efiacor, line 54: Might need some documentation for this here - https://porch.kpt.dev/docs/6_configuration_and_deployments/configurations/components/porch-controllers-config/#repository-controller-configuration

test/e2e/crd/lifecycle_test.go
```diff
@@ -201,7 +201,7 @@ var _ = Describe("Lifecycle", Ordered, Label("lifecycle"), func() {
 	// Known gap: dbPackageRevision.Delete only calls git delete for Published
 	// packages. Draft/proposed branches are not cleaned up. Tracked as Issue 36.
 
-	PIt("should clean up draft git branch when a Draft package is deleted", func() {
+	It("should clean up draft git branch when a Draft package is deleted", func() {
```
efiacor, line 204: Nice.

make/deploy.mk
```diff
@@ -52,6 +55,7 @@ run-in-kind: load-images-to-kind deployment-config deploy-current-config
 run-in-kind-v1alpha2: IMAGE_REPO=porch-kind## Build and deploy porch into a kind cluster with DB cache and v1alpha2 PackageRevision CRD creation
 run-in-kind-v1alpha2: PORCH_CACHE_TYPE=DB
 run-in-kind-v1alpha2: CREATE_V1ALPHA2_RPKG=true
+run-in-kind-v1alpha2: DB_PUSH_DRAFTS_TO_GIT=true
```
efiacor, line 58: Are we deploying with this enabled by default in CI?
Does it require a manual repo ctr sync for the e2e to run ok?

test/e2e/api/db_git_sync_test.go
```diff
@@ -0,0 +1,502 @@
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("updatePRR: PRR %q did not reflect update (keys %v) within %v", prName, resourceKeys, dbGitSyncWaitTimeout)
+	}
+}
+
+func (t *PorchSuite) triggerRepoSyncAndWaitForDraftBranch(repoName, giteaRepo, packageName, workspace string) string {
```
efiacor, line 58: In terms of UX, would the user need to understand that the repo ctr needs to have re synched for the draft/proposed branches to appear in git?
Is this acceptable UX?

test/e2e/suiteutils/gitea_test_utils.go
```diff
@@ -126,3 +143,287 @@ func RecreateGiteaRepo(t *testing.T, repoName string) {
 func (t *TestSuite) RecreateGiteaTestRepo() {
 	RecreateGiteaRepo(t.T(), PorchTestRepoName)
 }
+
```
efiacor, line 146: Why do we need all of the new manual Gitea funcs?

deployments/porch/9-controllers.yaml
rendre-greyling: Yep, going to need quite a bit of documentation updates. I can do it in this PR or a separate one, whichever is more convenient. Just wanted to get the implementation details squared away.

make/deploy.mk
rendre-greyling: I would prefer if we did, since it does not really have a performance impact anymore. No, afaik the only e2e tests which check for a draft branch in Gitea directly are the new ones I've added, and they manually trigger a sync of the repo. Not sure if I've missed any other tests that should be added for the repo ctr maybe?

test/e2e/api/db_git_sync_test.go
rendre-greyling: Yes, that should be documented. In terms of acceptable, I'm not sure and I guess it highly depends on the sync time of the repo ctr. 

The primary purpose is to ensure that draft branches can be restored from Git so that the porch DB does not have to be backed up. If said sync time is very long (for example 24 hours) and a draft is created in hour 1 and a disaster event happens at hour 23, that draft will be lost unless it is manually synced.

As I understand it, for most automated use cases, the PR is usually going to go from draft created, to published in a short span of time. So, the only time this really becomes an issue, and a draft branch has a chance of being pushed, is when creating PRs manually. 

So, to answer the question, is it acceptable, maybe? Unfortunately, I don't really see a different way of doing this right now. Option A is push on every change like we already had - with enough effort it can be made stable enough, but the performance is terrible. Option B is this. Option C is this, but decoupling this sync somehow from the repo controller and allowing it to run more often (seems truly terrible, convoluted, against the design of the repo ctr and makes performance worse with more requests to Git). 

Option D is a huge refactor involving Git simplification and we could consider moving the resources into the local Git repo instead of the DB and place the sync between the local Git repo and the external Git repo. Then we could push after every X commits to the PR. Simplifies the DB and could make the Git side easier to understand. Could even be a step towards eliminating the DB altogether if PRs are stored in etcd (of course if this is a direction the community wants to go in). This is something I'd like to explore when I have time and may even raise a PIP for it, but before Option D, we would need something that works.

Open to any other suggestions :)

test/e2e/suiteutils/gitea_test_utils.go
rendre-greyling: For the new e2e tests I needed to be able to interact with Gitea directly and delete or recreate PRs there to test all the functionality of the db sync.

deployments/porch/9-controllers.yaml
efiacor: Can be done in a follow up but try not to let it get lost

test/e2e/api/db_git_sync_test.go
efiacor: I think we may have use cases for long(er) lived drafts in git. @nagygergo @kushnaidu Could ye confirm?

efiacor: We go with a working solution for now with a view to improve later.

liamfallon (approved):

pkg/cache/dbcache/dbpackagerevision.go
```diff
@@ -408,7 +393,17 @@ func (pr *dbPackageRevision) SetMeta(ctx context.Context, meta metav1.ObjectMeta
 	_, span := tracer.Start(ctx, "dbPackageRevision::SetMeta", trace.WithAttributes())
 	defer span.End()
 
+	lockPkgKey(pr.pkgRevKey.PkgKey)
+	defer unlockPkgKey(pr.pkgRevKey.PkgKey)
```
Catalin-Stratulat-Ericsson, line 397: i presume this new mutex has been checked for race condition potential in your testing or using the vscode env var for porch server?

Catalin-Stratulat-Ericsson (approved):
great work

pkg/cache/dbcache/dbpackagerevision.go
rendre-greyling: Yep, tested it as much as possible for race conditions

